PR #86, opened 2023-07-03
SUP-15433: Fixed i18n input/select value propagation

deckdom:


cms-ui/libs/cms-components/src/lib/form-controls/i18n-input/i18n-input.component.ts
```diff
@@ -79,6 +79,7 @@ export class I18nInputComponent implements ControlValueAccessor, Validator, OnIn
                 // since the value is already present/updated.
                 const tmp = cloneDeep(this.i18nData || {});
                 tmp[this.language] = value;
+                this.i18nData = tmp;
```
deckdom, line 82: `this.i18nData[this.language] = value;`

PR #102, opened 2023-07-12
Add support for setting Global ID for Overviews from outside. 

plyhun:


cms-core/src/main/java/com/gentics/contentnode/rest/util/ModelBuilder.java
```diff
@@ -735,6 +736,12 @@ public static com.gentics.contentnode.rest.model.Overview getOverview(
 			Overview nodeOverview) throws NodeException {
 		com.gentics.contentnode.rest.model.Overview restOverview = new com.gentics.contentnode.rest.model.Overview();
 
+		if (nodeOverview == null) {
+			return restOverview;
+		}
+
+		// Fill IDs
+		restOverview.setGlobalId(nodeOverview.getGlobalId().toString());
```
npomaroli, line 744: Add null-check for nodeOverview.getGlobal()

npomaroli (changes requested):

npomaroli (approved):

PR #114, opened 2023-07-25
HTTP/2 serving support

plyhun:


cms-core/src/main/java/com/gentics/contentnode/runtime/ConfigurationValue.java
```diff
@@ -16,6 +16,11 @@ public enum ConfigurationValue {
 	 */
 	HTTP_PORT("HTTP_PORT", "com.gentics.contentnode.http.port", "server.port", () -> "8080"),
 
+	/**
+	 * Use HTTP/2 for connections
+	 */
+	USE_HTTP2("USE_HTTP2", "com.gentics.contentnode.http.useH2", "server.useH2", () -> "false"),
```
npomaroli, line 22: Change names:
Environment variable (and enum value): HTTP2
System property: "com.gentics.contentnode.http2"
Configuration value: "server.http2"

cms-oss-changelog/src/changelog/entries/2023/07/7505.GPU-1159.enhancement
```diff
@@ -0,0 +1 @@
+Support of serving over HTTP/2 (h2c) protocol has been added.
```
npomaroli, line 1: Also add documentation (at least in admin_migration_cmp8.textile)

npomaroli (changes requested):

PR #121, opened 2023-08-03
Implement property substitution for some attributes

npomaroli:


cms-core/src/main/java/com/gentics/contentnode/object/ContentRepository.java
```diff
@@ -153,12 +160,24 @@ public abstract class ContentRepository extends AbstractContentObject implements
...
+			switch(cr.getPasswordType()) {
+			case none:
 				nodeCR.setPassword(null);
+				nodeCR.setPasswordProperty(false);
+				break;
+			case property:
+				if (cr.getPasswordProperty() != null) {
+					nodeCR.setPassword(cr.getPasswordProperty());
+				}
+				nodeCR.setPasswordProperty(true);
```
npomaroli, line 173: Move into if-clause

cms-core/src/main/java/com/gentics/contentnode/object/ContentRepository.java
```diff
@@ -153,12 +160,24 @@ public abstract class ContentRepository extends AbstractContentObject implements
...
+				break;
+			case value:
+				if (cr.getPassword() != null) {
+					nodeCR.setPassword(cr.getPassword());
+				}
+				nodeCR.setPasswordProperty(false);
```
npomaroli, line 179: Move into if-clause

PR #136, opened 2023-09-06
GPU-1115: Manually resize page images on publish

plyhun:
Since we are aware of the usage of images hosted by the same project, as well as their manipulation properties, we can apply these manipulation in advance on the page publish, to make it being able to restrict the on-demand manipulation, which is resource consuming and is theoretically DDoS prone.


cms-core/src/main/java/com/gentics/contentnode/publish/mesh/MeshPublisher.java
```diff
@@ -3224,10 +3263,13 @@ public void handleRenderedEntries(boolean preview, int nodeId, Supplier<FieldMap
 							String string = ObjectTransformer.getString(o, null);
 							if (string != null) {
 								field.add(string);
+								gisChecker.accept(string);
 							}
 						}
 					} else {
-						fields.put(entry.getMapname(), new StringFieldImpl().setString(ObjectTransformer.getString(value, null)));
+						String string = ObjectTransformer.getString(value, null);
```
plyhun, line 3270: The main point so far is - if we collect the image manipulations per page, when and how to apply them eventually? We need to find those images in the first place, and then either postpone their publish until the very end with all the manipulations collected, or make an ad-hoc manipulations on each and every page.

PR #146, opened 2023-09-19
WIP: GPU-992 - Java 17 Update

guggi:


cms-js-lib/pom.xml
```diff
@@ -134,7 +134,7 @@
 					<execution>
 						<?m2e execute?>
 						<id>minify</id>
-						<phase>compile</phase>
+						<phase>none</phase>
```
guggi, line 137: todo: change back to compile or fix

PR #230, opened 2024-03-21
Hotfix 6.0.x sup 16004

plyhun:


cms-oss-server/src/main/java/com/gentics/contentnode/servlets/SelectedSymlinkAllowedResourceAliasChecker.java
```diff
@@ -0,0 +1,38 @@
...
+import org.eclipse.jetty.server.AllowedResourceAliasChecker;
+import org.eclipse.jetty.server.handler.ContextHandler;
+import org.eclipse.jetty.util.log.Log;
+import org.eclipse.jetty.util.log.Logger;
+
+/**
+ * An extension of {@link AllowedResourceAliasChecker} which will allow symlinks alias to arbitrary
+ * targets, if the path is in the whitelist.
+ */
+public class SelectedSymlinkAllowedResourceAliasChecker extends AllowedResourceAliasChecker
```
plyhun, line 17: Here we cannot reuse an existing `SymlinkAllowedResourceAliasChecker` for two reasons.
1. It is enabled by default, thus - useless.
2. It is useless, because of its ultimate dependency on a base webroot, which is set to `/webroot`, which is based on an application classpath root, and not the root we set in the distinct handler.

npomaroli (changes requested):
Tested with the example setup of the ticket and it did not work (static file from symlinked package is not availalbe).

plyhun:
> Tested with the example setup of the ticket and it did not work (static file from symlinked package is not availalbe).

I was able to replicate the issue on a local setup. Will re-check with the provided docker environment as well.

cms-oss-server/src/main/java/com/gentics/contentnode/servlets/SelectedSymlinkAllowedResourceAliasChecker.java
plyhun: NB: Fresher Jetty version extends the possibility of AliasCheckers, though it requires a refactoring / overall stability check.
https://github.com/jetty/jetty.project/issues/8259

npomaroli (approved):

PR #231, opened 2024-03-21
SUP-15443: Fix of content staging crash

plyhun:


cms-core/src/main/java/com/gentics/contentnode/staging/StagingUtil.java
```diff
@@ -80,7 +80,7 @@ public static <T extends StageableNodeObject, K> Map<K, StagingStatus> checkStag
 			return null;
 		}
 
-		StagingStatusService service = StreamSupport.stream(loader.spliterator(), false).findFirst().orElse(null);
+		StagingStatusService service = StreamSupport.stream(loader.spliterator(), false).findAny().orElse(null);
```
npomaroli, line 83: I don't understand how this would fix the exception, which was described in the ticket.
The only difference between findFirst() and findAny() is that findFirst() will always return the first element in the stream (if it is ordered), and findAny() can return any.
Did you actually reproduce the problem and can verify that this change is a fix for it?

plyhun: All I could find is a reference to the old declined JDK bug, which is still not reproducible in CMS OSS setup.
https://bugs.openjdk.org/browse/JDK-8230843
So I have improved the code anyway, since 1) it is indeed not exactly thread safe to call the service loader from the single thread-unsafe static member, 2) those calls simply don't need Streams.

PR #238, opened 2024-03-26
Exclusive READ(PUBLISHED) permissions for the published nodes.

plyhun:


cms-core/src/main/java/com/gentics/contentnode/publish/mesh/MeshPublisher.java
```diff
@@ -4979,11 +4981,12 @@ protected Completable setPermissions(Collection<String> roles) {
 					.collect(Collectors.toList()));
```
plyhun, line 4981: Shouldn't we restrict the project/branch permissions up above in the same way?

cms-core/src/test/java/com/gentics/contentnode/tests/publish/mesh/MeshPublishRolesTest.java
```diff
@@ -19,6 +19,23 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.fail;
 
+import java.util.ArrayList;
```
plyhun, line 22: So I've tried to test this explicit case: alter a permission for an existing page, and re-publish it. It does not work since the existence of `supportsPublishOnCreate`, as the permissions never get updated with this flag. This, however, may be a bug as well. See `MeshPublisher#setRolePermissions`.

npomaroli: The flag `supportsPublishOnCreate` only determines whether permissions can be set while creating/updating nodes, or it is required to do this in a separate step.
I added a test case for this issue: `MeshPublishRolesTest.testPagePermissionsWithModifiedRootNodePermissions()`

PR #245, opened 2024-04-02
SUP-16409: Move objects between nodes + Mesh Publish (UT)

plyhun:


cms-core/src/test/java/com/gentics/contentnode/tests/publish/mesh/MeshPublishTest.java
```diff
@@ -619,6 +622,95 @@ public void testPageFilenameConflictDelete() throws Exception {
...
+			return create(Folder.class, f -> {
+				f.setMotherId(otherNode.getFolder().getId());
+				f.setName("Testfolder 2");
+				f.setPublishDir("folder2");
+			});
+		});
+
+		// publish
+		try (Trx trx = new Trx()) {
+			context.publish(false);
```
plyhun, line 679: Crash here.

plyhun:
```
14:27:31.558 [] ERROR [vert.x-worker-thread-17] [c.g.m.g.OrientDBDatabase] - com.orientechnologies.orient.core.storage.ORecordDuplicatedException: Cannot index record #1122:5: found duplicated key 'a7e7596bf0f911ee96030242ac120002' in index 'MeshVertexImpl' previously assigned to the record #1121:5
	DB name="tinkerpop1712066498378" INDEX=MeshVertexImpl RID=#1121:5
14:27:31.558 [] ERROR [vert.x-eventloop-thread-13] [c.g.m.r.r.FailureHandler] - Error for request in path: /api/v1/othertestproject/nodes/a7e7596bf0f911ee96030242ac120002?branch=othertestproject&etag=false&fields=uuid,parent
14:27:31.558 [] ERROR [vert.x-eventloop-thread-13] [c.g.m.r.r.FailureHandler] - Error:
com.gentics.mesh.core.rest.error.GenericRestException: 500 Internal Server Error error_internal params {}
	at com.gentics.mesh.core.rest.error.Errors.error(Errors.java:90)
	at com.gentics.mesh.graphdb.OrientDBDatabase.tx(OrientDBDatabase.java:535)
	at com.gentics.mesh.core.verticle.handler.HandlerUtilities.syncTx(HandlerUtilities.java:345)
	at com.gentics.mesh.core.verticle.handler.HandlerUtilities.createOrUpdateElement(HandlerUtilities.java:189)
	at com.gentics.mesh.core.verticle.handler.HandlerUtilities.createOrUpdateElement(HandlerUtilities.java:172)
	at com.gentics.mesh.core.verticle.handler.HandlerUtilities.updateElement(HandlerUtilities.java:153)
	at com.gentics.mesh.core.endpoint.handler.AbstractCrudHandler.handleUpdate(AbstractCrudHandler.java:89)
	at com.gentics.mesh.core.endpoint.node.NodeEndpoint.lambda$addUpdateHandler$20(NodeEndpoint.java:524)
	at io.vertx.ext.web.impl.BlockingHandlerDecorator.lambda$handle$0(BlockingHandlerDecorator.java:48)
	at io.vertx.core.impl.ContextBase.lambda$null$0(ContextBase.java:137)
	at io.vertx.core.impl.ContextInternal.dispatch(ContextInternal.java:264)
	at io.vertx.core.impl.ContextBase.lambda$executeBlocking$1(ContextBase.java:135)
	at io.vertx.core.impl.TaskQueue.run(TaskQueue.java:76)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:829)
14:27:31.559 [] ERROR [vert.x-eventloop-thread-13] [i.v.e.w.h.i.LoggerHandlerImpl] - 172.17.0.1 - POST /api/v1/othertestproject/nodes/a7e7596bf0f911ee96030242ac120002?branch=othertestproject&etag=false&fields=uuid,parent HTTP/1.1 500 178 - 8 ms

Mesh output for failed test testUnrecoverableFilenameConflict(com.gentics.contentnode.tests.publish.mesh.MeshPublishTest)
```

plyhun: A line earlier
```
14:27:31.515 [] WARN  [vert.x-eventloop-thread-13] [i.v.e.w.h.i.LoggerHandlerImpl] - 172.17.0.1 - DELETE /api/v1/testproject/nodes/a7e7596bf0f911ee96030242ac120002/languages/de?recursive=true HTTP/1.1 404 220 - 2 ms
14:27:31.558 [] ERROR [vert.x-worker-thread-17] [c.g.m.g.OrientDBDatabase] - com.orientechnologies.orient.core.storage.ORecordDuplicatedException: Cannot index record #1122:5: found duplicated key 'a7e7596bf0f911ee96030242ac120002' in index 'MeshVertexImpl' previously assigned to the record #1121:5
	DB name="tinkerpop1712066498378" INDEX=MeshVertexImpl RID=#1121:5
14:27:31.558 [] ERROR [vert.x-eventloop-thread-13] [c.g.m.r.r.FailureHandler] - Error for request in path: /api/v1/othertestproject/nodes/a7e7596bf0f911ee96030242ac120002?branch=othertestproject&etag=false&fields=uuid,parent
```
shows that an attempt to delete an old node is performed, with no success.

cms-core/src/main/java/com/gentics/contentnode/factory/object/PageFactory.java
```diff
@@ -2515,12 +2515,6 @@ public OpResult move(Folder target, int targetChannelId, boolean allLanguages) t
 					List<ContentLanguage> targetNodeLanguages = target.getOwningNode().getLanguages();
 					pageLanguages.removeAll(targetNodeLanguages);
 
-					// when the page to move does not have a language, but the target node has languages, we let the page have the first
```
plyhun, line 2518: This part of the code causes the UT to fail, as initially the page being moved does not have a language, and so I have doubts that we need to set those forcefully on a move.

cms-core/src/main/java/com/gentics/contentnode/publish/mesh/MeshPublisher.java
```diff
@@ -2133,18 +2133,20 @@ protected boolean checkObjectConsistency(MeshProject project, VersioningParamete
 		for (int objectType : types) {
 			Class<? extends NodeObject> clazz = t.getClass(objectType);
 
-			Map<Integer, Set<String>> map = PublishQueue.getObjectIdsWithAttributes(clazz, true, checkedNode, Action.DELETE, Action.REMOVE, Action.OFFLINE);
+			Map<Integer, Set<String>> deleteOfflineMap = PublishQueue.getObjectIdsWithAttributes(clazz, true, checkedNode, Action.DELETE, Action.OFFLINE);
```
plyhun, line 2136: Here we tell apart the DELETEd and REMOVEd objects just as documented - DELETE means complete removal, REMOVE means the object being moved elsewhere. Forms require this tellapart explicitly.

pom.xml
```diff
@@ -102,6 +102,7 @@
 		<surefire.forkcount.mesh>4</surefire.forkcount.mesh>
 		<!-- Since we have some glitch tests, we rerun failing tests once -->
 		<surefire.rerunFailingTestsCount>1</surefire.rerunFailingTestsCount>
+		<surefire.skip>false</surefire.skip>
```
plyhun, line 105: I'd like to have an ability to build the whole project, including unit tests, without an actual run of the tests.

cms-core/src/main/java/com/gentics/contentnode/factory/object/PageFactory.java
npomaroli: This part must not be removed, it was added on purpose (the reason is in the comment).
But the failing UT shows another problem: When moving a page from Node A to Node B, all language variants of the page must be removed from Mesh (for Node A), not only the language variant, the page currently has. I will provide a fix for this.

cms-core/src/main/java/com/gentics/contentnode/publish/mesh/MeshPublisher.java
```diff
@@ -2155,29 +2157,46 @@ protected boolean checkObjectConsistency(MeshProject project, VersioningParamete
...
-				}
-				if (meshUuid != null) {
-					toDelete.computeIfAbsent(meshUuid, k -> new HashSet<>()).add(meshLanguage);
 				}
 			}
+			if (controller.publishProcess && (PublishController.getState() != PublishController.State.running)) {
+				logger.debug(String.format("Stop checking offline objects, because publisher state is %s", PublishController.getState()));
+				return false;
+			}
+			Stream.of(deleteOfflineMap.entrySet().stream(), removeMap.entrySet().stream())
```
npomaroli, line 2182: Doing it this way moves the check for stopped publish processes out of the loop. This must not be done, because then the publish process cannot be interrupted during removal of objects from mesh.

cms-oss-changelog/src/changelog/entries/2024/04/7736.SUP-16409.bugfix
```diff
@@ -0,0 +1 @@
+When moving an object (Page, File etc) between the nodes, that point to the completely different Mesh CRs, the following publish process may fail with an item duplication error. This has been fixed.
```
npomaroli, line 1: Actually, the problem of duplicate items only arises when objects are moved between nodes, which publish into the same Mesh instance (via the same or different CRs).

npomaroli (changes requested):

pom.xml
plyhun: `-DskipTests` skips building test deps, which means that every new build, that depends on those, fails. I need them to be built, but not run.

plyhun: > I have verified that this is in fact the case.

It has been never the case. For instance, if building CMS OSS with this flag, the following build of CMS fails on missing parent test-lib, which has not been build, thus - not installed.

PR #254, opened 2024-04-10
Fix handling of server version

npomaroli:
Fix URLs to changelog and guides
Remove link to manual

### DIFF
--- a/cms-core/src/main/java/com/gentics/contentnode/version/ServerVariantService.java
+++ b/cms-core/src/main/java/com/gentics/contentnode/version/ServerVariantService.java
@@ -2,10 +2,27 @@ package com.gentics.contentnode.version;
 
 import com.gentics.contentnode.update.CMSVersion;
 
+/**
+ * Interface for service, that returns CMS variant specific information
+ */
 public interface ServerVariantService {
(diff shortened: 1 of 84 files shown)


cms-core/src/main/java/com/gentics/contentnode/version/ServerVariantService.java
```diff
@@ -0,0 +1,11 @@
+package com.gentics.contentnode.version;
+
+import com.gentics.contentnode.update.CMSVersion;
+
+public interface ServerVariantService {
```
npomaroli, line 5: Javadoc

cms-restapi/src/main/java/com/gentics/contentnode/rest/model/response/admin/Update.java
```diff
@@ -0,0 +1,29 @@
+package com.gentics.contentnode.rest.model.response.admin;
+
+import java.io.Serializable;
+
+public class Update implements Serializable {
```
npomaroli, line 5: Javadoc

npomaroli (commented):
Javadoc!!!

PR #327, opened 2024-07-08
Use Maven 3.8.8 instead of 3.9.x

npomaroli:


yrucrem:
No longer relevant after switch to nexus.

PR #328, opened 2024-07-09
GPU-1560: UI Integration Tests

deckdom:


cms-ui/apps/admin-ui-e2e/src/fixtures/auth.json
```diff
@@ -1,4 +1,8 @@
 {
+    "setup": {
```
deckdom, line 2: Can be removed

PR #346, opened 2024-08-09
Fix permission check for images

npomaroli:


cms-ui/apps/editor-ui/src/app/core/providers/permissions/permission.service.ts
```diff
@@ -802,14 +802,32 @@ export class PermissionService {
...
-                create: perm.createitems,
-                delete: perm.deleteitems,
-                edit: perm.updateitems,
-                import: perm.importitems,
-                localize: perm.createitems,
-                upload: perm.createitems,
-                unlocalize: perm.createitems,
-                view: perm.readitems,
+                create:
+                    (!!perm && perm.createitems) ||
```
deckdom, line 806: Null coalescing would be better/more clear: `perm?permcreateitems ?? role.file.createitems`

deckdom (approved):

cms-ui/apps/editor-ui/src/app/core/providers/permissions/permission.service.ts
npomaroli: I copied this from the file permissions below.

PR #347, opened 2024-08-12
SUP-15430: Fix spellchecker

guggi:
Issue: https://jira-gentics.atlassian.net/browse/SUP-15430

### DIFF
--- a/cms-core/src/main/resources/com/gentics/contentnode/runtime/config.yml
+++ b/cms-core/src/main/resources/com/gentics/contentnode/runtime/config.yml
@@ -114,6 +114,7 @@ aloha_plugins_global:
   - "common/commands"
   - "gcn/gcn"
   - "common/dom-to-xhtml"
+  - "gcn/spellcheck"
 aloha_plugins_node: []
 http_auth_login: ""
 fileupload_manipulator_url: ""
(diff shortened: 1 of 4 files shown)


cms-core/src/main/resources/com/gentics/contentnode/runtime/config.yml
```diff
@@ -114,6 +114,7 @@ aloha_plugins_global:
   - "common/commands"
   - "gcn/gcn"
   - "common/dom-to-xhtml"
+  - "gcn/spellcheck"
```
guggi, line 117: Maybe remove this to not have it activated as default?

deckdom: As far as I can tell, this isn't a special/paid/enterprise feature, and spellchecks are quite common, so having this enabled on default should be fine.

Regarding the change in the Aloha-Editor Repo:
Please add the entry for the `gcnSpellCheck` in here instead (Line `78`):

```yml
      - id: view
        merge:
          components: 'append'
        components:
          - 'gcnSpellCheck'
```

This should simply add the entry to the settings

deckdom (changes requested):
Please see comment

cms-core/src/main/resources/com/gentics/contentnode/runtime/config.yml
guggi: b57a6e7d

PR #353, opened 2024-08-26
WIP: AdminResource: Feature manipulation

plyhun:
GPU-1792

### DIFF
--- a/cms-core/src/test/java/com/gentics/contentnode/tests/rest/admin/AdminResourceFeatureManagementTest.java
+++ b/cms-core/src/test/java/com/gentics/contentnode/tests/rest/admin/AdminResourceFeatureManagementTest.java
@@ -1,0 +1,100 @@
+package com.gentics.contentnode.tests.rest.admin;
+
+import static com.gentics.contentnode.factory.Trx.*;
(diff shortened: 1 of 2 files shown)


cms-core/src/test/java/com/gentics/contentnode/tests/rest/admin/AdminResourceFeatureManagementTest.java
```diff
@@ -0,0 +1,100 @@
...
+			Transaction t = TransactionManager.getCurrentTransaction();
+			NodePreferences prefs = t.getNodeConfig().getDefaultPreferences();
+			return prefs.getFeature(Feature.TAG_IMAGE_RESIZER.getName());
+		});
+		assertThat(enabled).isTrue();
+	}
+
+	@Test
+	public void testFeatureWebpConversion() throws NodeException {
+//		operate(() -> {
```
plyhun, line 76: This test fails. CMS checks the general availability of the node-specific feature, before it checks its availability on the requested node, so the check returns true, even if it is removed from the node. There must be another mechanism of checking the feature availability on the node...

PR #364, opened 2024-09-16
GPU-1606: UI

deckdom:


deckdom (commented):
1. Form-Properties Form -> Rename
2. Request Models update (Use Editable<ITEM>Props Types)
3. Page properties is disabled on default -> check why

PR #365, opened 2024-09-18
SUP-17445: Fix loading of styles for custom tag editors

guggi:
Issue: https://jira-gentics.atlassian.net/browse/SUP-17445


guggi:
Created a new one with the correct base

PR #385, opened 2024-10-21
SUP-17472: Added integration tests

deckdom:


cms-ui/apps/editor-ui-e2e/src/e2e/page-management.cy.ts
```diff
@@ -131,6 +140,29 @@ describe('Page Management', () => {
             });
         });
 
+        it.only('should have the testing object-property category open on default', () => {
```
deckdom, line 143: Remove `.only`

PR #394, opened 2024-10-22
SUP-17299: Mesh Browser loading indicators & general cleanup

deckdom:


deckdom:
Do performance checks with a lot of schemas and nodes

PR #398, opened 2024-10-25
SUP-17208: Correctly process channel languages

plyhun:


npomaroli (changes requested):
When creating a channel, the UI still lets you configure languages in the "languages" tab and if languages are added, you get error messages now (that the operation is not allowed for channels).

npomaroli (approved):

PR #449, opened 2025-01-17
SUP-17941: Fix language specific permissions

npomaroli:
Users need "edit" permission for publishing page (into queue)


cms-ui/apps/editor-ui/src/app/content-frame/components/aloha-attribute-button-renderer/aloha-attribute-button-renderer.component.ts
```diff
@@ -113,7 +113,7 @@ export class AlohaAttributeButtonRendererComponent
             })
             .catch(err => {
                 this.inputDropdown = null;
-                if (err instanceof ModalCloseError && err.reason !== ModalClosingReason.ERROR) {
+                if (!wasClosedByUser(err)) {
```
npomaroli, line 116: not not

PR #470, opened 2025-02-06
SUP-17792: Show result breadcrumbs for GCMS search

plyhun:


cms-ui/apps/editor-ui/src/app/shared/providers/query-assembler/query-assembler-gcmssearch/query-assembler-gcmssearch.service.ts
```diff
@@ -327,18 +327,22 @@ export class QueryAssemblerGCMSSearchService {
...
                             isValidFilterForProperty = GCMSSEARCH_AVAILABLE_FILTERS_PAGE.some(p => p === paramfilterKey);
+                            mappedOptions.folder = true;
                             break;
                         case 'form':
                             isValidFilterForProperty = GCMSSEARCH_AVAILABLE_FILTERS_FORM.some(p => p === paramfilterKey);
+                            mappedOptions.folder = true;
                             break;
                         case 'folder':
                             isValidFilterForProperty = GCMSSEARCH_AVAILABLE_FILTERS_FOLDER.some(p => p === paramfilterKey);
                             break;
```
plyhun, line 338: No `folder` options set for the folders call, since here it looks for the subfolders, which is opposite from what we need. The folder search' breadcrumbs are already in place though.

deckdom (approved):

PR #471, opened 2025-02-06
SUP-17402: Fix the display properties sorting

plyhun:


cms-ui/apps/editor-ui/src/app/shared/components/display-field-selector/display-field-selector.component.tpl.html
```diff
@@ -2,16 +2,15 @@
     <h4>{{ 'modal.display_field_selector_title' | i18n }}</h4>
 </div>
 <div class="modal-content">
+    <gtx-contents-list-item *ngIf="type !== 'folder'">
```
plyhun, line 5: As the unsortable `path` item was a part of a sortable-list, the sorting function was getting the wrong indices to move the item from/to.

PR #480, opened 2025-02-11
SUP-17784 Fix RSA login

yrucrem:


cms-core/src/main/java/com/gentics/contentnode/etc/LoginService.java
```diff
@@ -1,19 +1,54 @@
...
+	default boolean isDefaultService() {
+		return false;
+	}
+
+	/**
+	 * Determine whether the secure flag of the session cookie should be set.
+	 * @return true for secure, false otherwise
+	 */
+	static boolean isCookieSecure() throws TransactionException {
+		return TransactionManager.getCurrentTransaction().getNodeConfig().getDefaultPreferences().getFeature("secure_cookie");
```
npomaroli, line 51: Use NodeConfigRuntimeConfiguration instead of transaction

npomaroli (changes requested):

cms-core/src/main/java/com/gentics/contentnode/etc/LoginService.java
yrucrem: Done.

PR #483, opened 2025-02-14
SUP-18072: Extend the number of allowed symbols of datasource entry value from 50 to 16383

plyhun:


cms-ui/apps/admin-ui/src/app/features/data-source/components/data-source-entry-properties/data-source-entry-properties.component.html
```diff
@@ -12,16 +12,14 @@
...
         ></gtx-input>
-    </div>
 
-    <div class="columns small-6">
-        <gtx-input
+		<gtx-textarea
             type="text"
             [required]="true"
-            maxlength="50"
+            maxlength="16383"
```
plyhun, line 19: 65535 max MariaDB `TEXT` / 4 = worst UTF-8 case.

PR #484, opened 2025-02-14
SUP-18084: Less secure Mesh token clipboard copy

plyhun:
Not sure if this has to be added to the changelog, giving its "solution"...

### DIFF
--- a/cms-ui/apps/admin-ui/src/app/mesh/components/copy-value/copy-value.component.html
+++ b/cms-ui/apps/admin-ui/src/app/mesh/components/copy-value/copy-value.component.html
@@ -1,9 +1,14 @@
 <div *ngIf="label" class="label">{{ label }}</div>
-<div class="wrapper" [class.disabled]="disabled" (copy)="cancelEvent($event)">
+<div class="wrapper" [class.disabled]="disabled" (copy)="cancelIfDisabled($event)">
     <div class="content-wrapper">
         <div class="content">{{ value }}</div>
     </div>
-    <button class="copy-button" [disabled]="disabled" [title]="'shared.copy' | i18n" (click)="copyToClipboard()">
+    <button
+        class="copy-button"
+        [disabled]="disabled"
+        [title]="'shared.copy' | i18n"
+        (click)="copyToClipboard()"
+    >
         <icon>{{ icon }}</icon>
     </button>
 </div>
--- a/cms-ui/apps/admin-ui/src/app/mesh/components/copy-value/copy-value.component.html
+++ b/cms-ui/apps/admin-ui/src/app/mesh/components/copy-value/copy-value.component.html
@@ -3,7 +3,7 @@
     <div class="content-wrapper">
         <div class="content">{{ value }}</div>
     </div>
-    <button class="copy-button" [disabled]="disabled" [title]="'shared.copy' | i18n" (click)="copyToClipboard()">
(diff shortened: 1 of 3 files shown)


cms-ui/apps/admin-ui/src/app/mesh/components/copy-value/copy-value.component.html
```diff
@@ -3,7 +3,7 @@
     <div class="content-wrapper">
         <div class="content">{{ value }}</div>
     </div>
-    <button class="copy-button" [disabled]="disabled" [title]="'shared.copy' | i18n" (click)="copyToClipboard()">
+    <button *ngIf="clipboardAllowed | async" class="copy-button" [disabled]="disabled" [title]="'shared.copy' | i18n" (click)="copyToClipboard()">
```
deckdom, line 6: 1. Please avoid `async` as much as possible
2. This doesn't even do anything, as `clipboardAllowed` is always `true`

cms-ui/apps/admin-ui/src/app/mesh/components/copy-value/copy-value.component.ts
```diff
@@ -35,17 +35,46 @@ export class CopyValueComponent extends BaseComponent {
         super(changeDetector);
     }
 
-    async copyToClipboard(): Promise<void> {
+    copyToClipboard(): Promise<void> {
```
deckdom, line 38: It should first attempt to do the secure way, and use the insecure way as fallback instead of being the default solution.

cms-ui/apps/admin-ui/src/app/mesh/components/copy-value/copy-value.component.ts
```diff
@@ -35,17 +35,46 @@ export class CopyValueComponent extends BaseComponent {
...
             if (this.animate) {
                 this.animateCopy(1)
             }
         } catch (err) {
             this.copy.emit(err);
         }
+		return Promise.resolve();
     }
+	// https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1245
+	/*  async copyToClipboard(): Promise<void> {
```
deckdom, line 61: Remove unused code instead of commenting it out - See comment above however, shouldn't be fully "unused"

deckdom:
Missing: Changelog entry

deckdom (changes requested):

cms-ui/apps/admin-ui/src/app/mesh/components/copy-value/copy-value.component.ts
plyhun: This is the point -- an issue in DOM lib prevents the `clipboardAllowed()` from being succeeded. The only way we can check that the 'official' clipboard functionality works is to try it out, expecting the exception to be thrown in case of an error. This happens not all the time, for instance, at my local insecure Firefox 135.0 setup. This is why I left the official check commented and use the hack all the time, until the TS issue is fixed.

cms-ui/apps/admin-ui/src/app/mesh/components/copy-value/copy-value.component.html
plyhun: Here it has to be a call to an async API function, so it _would have been_ unavoidable, if it had worked (see a TS issue reference below).

PR #485, opened 2025-02-14
SUP-18083: Log startup config files better

plyhun:


cms-core/src/main/java/com/gentics/contentnode/runtime/NodeConfigRuntimeConfiguration.java
```diff
@@ -303,14 +303,22 @@ public static Map<String, Object> loadConfiguration() throws NodeException {
 						// merge all files
 						for (File subConfigFile : configurationFiles) {
 							runtimeLog.info(String.format("Reading configuration from %s", subConfigFile.getAbsolutePath()));
-							data = merge(data, mapper.readValue(subConfigFile, Map.class));
+							try {
+								data = merge(data, mapper.readValue(subConfigFile, Map.class));
```
npomaroli, line 307: Why do we not also check via
if (!mapper.readTree(subConfigFile).isEmpty()) {
whether the file is empty?

npomaroli (changes requested):

cms-core/src/main/java/com/gentics/contentnode/runtime/NodeConfigRuntimeConfiguration.java
plyhun: Cause it might be expensive, depending on a config file contents.

npomaroli: The CMS does not have that many configuration values to make parsing the config files too expensive.
And we can avoid reading the same file twice by using
mapper.treeToValue() after mapper.readTree()

plyhun:
Test cases added.

cms-core/src/main/java/com/gentics/contentnode/runtime/NodeConfigRuntimeConfiguration.java
```diff
@@ -268,6 +275,24 @@ public static void overwrite(Map<String, Object> config) {
 	public static Map<String, Object> loadConfiguration() throws NodeException {
 		// load data
 		ObjectMapper mapper = new ObjectMapper(new YAMLFactory()).setDefaultMergeable(true);
+		// customize the mapper: on unknown value type, improper format or improper file contents return null, no exception
```
npomaroli, line 278: Why this?
We still want parsing malformed configuration files to fail with an exception.

cms-core/src/main/java/com/gentics/contentnode/rest/util/MiscUtils.java
```diff
@@ -1632,25 +1632,42 @@ public boolean canResolve() {
...
 	@SuppressWarnings("unchecked")
-	public static Map<String, Object> merge(Map<String, Object> map, Map<String, Object> other) {
-		for (Map.Entry<String, Object> e : other.entrySet()) {
-			if (e.getValue() == null) {
-				map.put(e.getKey(), null);
-			} else {
-				map.merge(e.getKey(), e.getValue(), (oldVal, newVal) -> {
-					if (oldVal instanceof Map && newVal instanceof Map) {
-						return merge((Map<String, Object>) oldVal, (Map<String, Object>) newVal);
+	public static Map<String, Object> merge(Map<String, Object> map, Map<String, Object> other, boolean mergeNullValues) {
```
npomaroli, line 1653: Why this change?
This has nothing to do with skipping empty configuration files.

npomaroli (changes requested):
Please keep the change simple and clean. All we want is:
1. reading the configuration should not fail on empty files (or files only containing comments)
2. when reading fails on a malformatted configuration file, the exception should contain the name of that file

cms-core/src/main/java/com/gentics/contentnode/runtime/NodeConfigRuntimeConfiguration.java
plyhun: Eh. Old attempt to deal with the malformed configs.

npomaroli (approved):

PR #489, opened 2025-02-17
SUP-17875: Fix the field pickers in the Mesh schema editor

plyhun:


cms-ui/apps/admin-ui/src/app/mesh/components/schema-properties/schema-properties.component.ts
```diff
@@ -59,8 +62,8 @@ export class SchemaPropertiesComponent extends BasePropertiesComponent<EditableS
         });
     }
 
-    protected configureForm(_value: EditableSchemaProperties, _loud?: boolean): void {
-        // no-op
+    protected configureForm(value: EditableSchemaProperties, _loud?: boolean): void {
+        this.fieldNames = value?.fields?.map?.(field => field.name);
```
plyhun, line 66: IDK if this is the best option to fill the options, but at least we know that neither of a control constructor or whitelist filter are called in time, e.g. when an actual value is obtained.

PR #494, opened 2025-02-19
SUP-17215: Fix the scheduler status widget

plyhun:


cms-ui/apps/admin-ui/src/app/shared/components/widget-publishing-process/widget-publishing-process.component.ts
```diff
@@ -64,9 +64,9 @@ export class WidgetPublishingProcessComponent extends SidebarItemComponent imple
 
         this.subscriptions.push(intervall$.pipe(
             startWith(0),
-            switchMap(() => this.adminOps.getJobs({failed: true})),
+            switchMap(() => this.schedulerOps.getAll({pageSize: 1, sort: '-edate'})),
```
plyhun, line 67: IDK if I need to pick more that one last execution. Perhaps this can be changed to last X entries.

deckdom: I don't think that checking it like this would display the correct status.
This will load all schedules, and checks for any failures in any of these schedules, which doesn't have to be the publish process.
While this was also done like this in the old implementation, it should be determined if this is actually what we want.

I'd suggest a proper endpoint or similar to get the latest publish information, as we can't determine the schedules to check (Technically we could by loading all schedules and tasks, checking which task is executing the publish etc, but that's insane overhead for something like this).

Please check with NOP when he got time.

npomaroli: There is a misunderstanding. The status label says "Last status of scheduler tasks" and that is exactly what it is supposed to show. So it's about all scheduler tasks (actually schedules) not only about the publish process.
I would propose to add a specific filtering mechanism to the endpoint /scheduler/schedule like we had for the previously used /scheduler/jobs (which was for the old PHP based scheduler) to return only failed/successful and/or active/inactive schedules.

cms-core/src/main/java/com/gentics/contentnode/rest/resource/impl/scheduler/SchedulerResourceImpl.java
```diff
@@ -371,6 +358,21 @@ public ScheduleListResponse listSchedules(
...
+					.filter(o -> isFailStatusFiltered(o, jobFilter, oo -> {
+						ScheduleModel model = scheduleModels.get(oo.getId());
+						if (model == null) {
+							// TODO does a schedule with no corresponding info equal to a failed execution?
+							return false;
+						}
+						ExecutionModel lastExecution = model.getLastExecution();
+						if (lastExecution == null) {
+							// Never executed schedule does not mean it is failed
+							return true;
```
npomaroli, line 370: Return value is the "fail" status (true when failed), so we should return false here

cms-core/src/main/java/com/gentics/contentnode/rest/resource/impl/scheduler/SchedulerResourceImpl.java
```diff
@@ -371,6 +358,21 @@ public ScheduleListResponse listSchedules(
...
+						}
+						// So as an execution with no result (yet)
+						return (lastExecution.getResult() == null || lastExecution.getResult());
```
npomaroli, line 373: Like before, the return value is reversed here

cms-core/src/test/java/com/gentics/contentnode/tests/scheduler/SchedulerResourceImplTest.java
```diff
@@ -63,6 +63,34 @@ public void givenScheduleWithEmbeddedTaskRequest_shouldHaveEmbeddedTaskInRespons
...
 	}
 
+	@Test
+	public void givenSchedule_filterByActive_shouldHaveResponse()
+			throws Exception {
+		SchedulerJobFilterParameterBean filter = new SchedulerJobFilterParameterBean();
+
+		filter.active = true;
+		ScheduleListResponse scheduleListResponse = new SchedulerResourceImpl().listSchedules(null, null, null, null, null, filter);
+		assertThat(scheduleListResponse.getItems()).asList().allMatch(a -> ((ScheduleModel) a).getActive());
```
npomaroli, line 73: When importing org.assertj.core.api.Assertions.assertThat instead of org.assertj.core.api.AssertionsForClassTypes.assertThat the cast in allMatch is not necessary.

cms-core/src/test/java/com/gentics/contentnode/tests/scheduler/SchedulerResourceImplTest.java
```diff
@@ -63,6 +63,34 @@ public void givenScheduleWithEmbeddedTaskRequest_shouldHaveEmbeddedTaskInRespons
...
+
+		filter.active = false;
+		scheduleListResponse = new SchedulerResourceImpl().listSchedules(null, null, null, null, null, filter);
+		assertThat(scheduleListResponse.getItems()).asList().allMatch(a -> !((ScheduleModel) a).getActive());
+	}
+
+	@Test
+	public void givenSchedule_filterByFailed_shouldHaveResponse()
```
npomaroli, line 81: This test does not test anything, because the response does not contain any items.
Generally, the test class should be modified so that each test creates its own test data and the test data is cleared before every test.

npomaroli (changes requested):

cms-core/src/test/java/com/gentics/contentnode/tests/scheduler/SchedulerResourceImplTest.java
```diff
@@ -63,6 +63,40 @@ public void givenScheduleWithEmbeddedTaskRequest_shouldHaveEmbeddedTaskInRespons
...
+			throws Exception {
+		SchedulerTask createdTask = givenTask();
+		givenScheduleWithTask(createdTask);
+
+		SchedulerJobFilterParameterBean filter = new SchedulerJobFilterParameterBean();
+
+		filter.failed = true;
+		ScheduleListResponse scheduleListResponse = new SchedulerResourceImpl().listSchedules(null, null, null, null, null, filter);
+		assertThat(scheduleListResponse.getItems()).isEmpty();;
```
npomaroli, line 93: This is still no meaningful test.
Please create test data for the following cases:
* schedule without execution
* schedule with failed last execution
* schedule with successful last execution

cms-core/src/test/java/com/gentics/contentnode/tests/scheduler/SchedulerResourceImplTest.java
```diff
@@ -63,6 +63,40 @@ public void givenScheduleWithEmbeddedTaskRequest_shouldHaveEmbeddedTaskInRespons
...
+
+		SchedulerJobFilterParameterBean filter = new SchedulerJobFilterParameterBean();
+
+		filter.active = true;
+		ScheduleListResponse scheduleListResponse = new SchedulerResourceImpl().listSchedules(null, null, null, null, null, filter);
+		assertThat(scheduleListResponse.getItems()).isNotEmpty().allMatch(a -> ((ScheduleModel) a).getActive());
+
+		filter.active = false;
+		scheduleListResponse = new SchedulerResourceImpl().listSchedules(null, null, null, null, null, filter);
+		assertThat(scheduleListResponse.getItems()).isEmpty();;
```
npomaroli, line 80: Not a meaningful test.
Please create test data for the following cases:
* active schedule
* inactive schedule

npomaroli (changes requested):

cms-ui/libs/cms-models/src/lib/models/request.ts
```diff
@@ -774,21 +774,23 @@ export interface ElasticSearchTypeSearchOptions {
     wastebin?: 'exclude' | 'include' | 'only';
 }
 
-export interface JobListRequestOptions extends BaseListOptionsWithPaging<Jobs> {
-    /**
-     * If set to true, only active jobs will be returned.
-     * If set to false, only inactive jobs will be returned. If not set (default), all jobs will be returned.
-     */
-    failed?: boolean;
+export interface SchedulableTaskListOptions {
```
deckdom, line 777: Please correct the JSDoc - They are swapped for some reason

deckdom (approved):
Approved with one nitpick, which would be neat if fixed

PR #495, opened 2025-02-19
SUP-15751: Get the tagmap entry objtypes back

plyhun:


cms-ui/apps/admin-ui/src/app/shared/components/tag-map-entry-table/tag-map-entry-table.component.ts
```diff
@@ -68,9 +68,9 @@ export class TagMapEntryTableComponent
             sortable: true,
         },
         {
-            id: 'object',
+            id: 'objType',
```
plyhun, line 71: Perhaps this rename isn't necessary, but I'd keep this similar to the backend DTOs.

deckdom (approved):

PR #502, opened 2025-02-24
SUP-18134: Remove discontinued spellcheck plugin from Aloha guides

plyhun:


cms-core/src/main/resources/com/gentics/contentnode/runtime/config.yml
```diff
@@ -78,8 +78,6 @@ aloha_settings:
       - id: view
         merge: 
           components: 'append' 
-        components:
```
deckdom, line 81: Please keep the default configuration for this plugin however.
If the plugin is not enabled, the button will also not be visible in the UI.
Otherwise we'd also need to update the documentation to add this back as well, so this seems easier.

deckdom (approved):

PR #525, opened 2025-03-19
GPU-1908: Protocol in the hostname

plyhun:


cms-core/src/main/java/com/gentics/contentnode/rest/resource/impl/NodeResourceImpl.java
```diff
@@ -899,12 +899,14 @@ public NodeLoadResponse add(NodeSaveRequest request) throws NodeException {
...
+			newNode.setFolder(rootFolder);			
 			newNode.setHostnameProperty(reqNode.getHostProperty());
 
-			if (reqNode.isHttps() != null) {
-				newNode.setHttps(reqNode.isHttps());
+			if (reqNode.hostnameContainsProtocol()) {
+				newNode.setHttps(reqNode.getHost().startsWith("https://"));
+				newNode.setHostname(reqNode.getHost().substring(reqNode.getHost().indexOf("://") + 3));
+			} else {
+				newNode.setHostname(reqNode.getHost());
```
npomaroli, line 909: This probably overwrites the host, which was extracted from the hostname property with null

cms-core/src/main/java/com/gentics/contentnode/rest/resource/impl/NodeResourceImpl.java
```diff
@@ -899,12 +899,14 @@ public NodeLoadResponse add(NodeSaveRequest request) throws NodeException {
...
 			}
 
-			newNode.setFolder(rootFolder);
-			newNode.setHostname(reqNode.getHost());
+			newNode.setFolder(rootFolder);			
 			newNode.setHostnameProperty(reqNode.getHostProperty());
 
-			if (reqNode.isHttps() != null) {
-				newNode.setHttps(reqNode.isHttps());
+			if (reqNode.hostnameContainsProtocol()) {
```
npomaroli, line 905: This also needs to be done when setting the hostname property (in the implementation EditableFactoryNode.setHostnameProperty(String))

cms-core/src/test/java/com/gentics/contentnode/tests/rest/node/NodeCreationTest.java
```diff
@@ -28,21 +29,101 @@ public class NodeCreationTest {
 	 * @throws NodeException
 	 */
 	@Test
-	public void testCreateWithPubDirSegment() throws NodeException {
+	public void testCreateForceInsecure() throws NodeException {
```
npomaroli, line 32: Also add tests for creating/updating nodes with a hostname property (with/without protocol)

cms-core/src/main/java/com/gentics/contentnode/factory/object/FolderFactory.java
```diff
@@ -6273,7 +6274,15 @@ public void resolveHostnameProperty() throws ReadOnlyException {
 			if (!org.apache.commons.lang3.StringUtils.isBlank(this.hostProperty)) {
 				String resolvedHost = substituteSingleProperty(this.hostProperty, Node.NODE_HOST_FILTER);
 				if (!StringUtils.isEqual(this.host, resolvedHost)) {
```
npomaroli, line 6276: This check is not correct, since this.host will only contain the hostname, but resolvedHost might contain the protocol also. We only want to set this.modified to true, if somehing (this.host or this.https) is actually changed. Better remove this check and use setHttps and setHostname as setter functions for MiscUtils.setHostAndProtocol

cms-core/src/main/java/com/gentics/contentnode/rest/resource/impl/NodeResourceImpl.java
```diff
@@ -1094,12 +1092,12 @@ public GenericResponse update(@PathParam("id") String nodeId, NodeSaveRequest re
 			if (!ObjectTransformer.isEmpty(reqNode.getHost())) {
```
npomaroli, line 1092: We don't want to do this, since reqNode.getHost() could contain the protocol

cms-core/src/main/java/com/gentics/contentnode/factory/object/FolderFactory.java
```diff
@@ -6273,7 +6274,15 @@ public void resolveHostnameProperty() throws ReadOnlyException {
...
-					this.host = resolvedHost;
+					try {
+						MiscUtils.setHostAndProtocol(resolvedHost, https -> this.https = https, host -> this.host = host);
+					} catch (NodeException e) {
```
npomaroli, line 6279: This is both ugly and dangerous, because throwing an IllegalStateException would cause an internal server error. Better change the method signature to throw NodeException instead of ReadOnlyException.

cms-core/src/main/java/com/gentics/contentnode/rest/util/MiscUtils.java
```diff
@@ -2475,4 +2475,34 @@ public static <I, O> O execOrNull(Function<I, O> function, I input) {
...
+	/**
+	 * Set the hostname and secure connection flag, based on an input string, that can possibly contain and URL, prefixed with a protocol.
+	 * 
+	 * @param input
+	 * @param secureSetter
+	 * @param hostSetter
+	 * @throws NodeException 
+	 */
+	public static void setHostAndProtocol(String input, Consumer<Boolean> secureSetter, Consumer<String> hostSetter) throws NodeException {
+		if (input != null) {
```
npomaroli, line 2498: Check for not empty

cms-core/src/test/java/com/gentics/contentnode/tests/rest/node/NodeCreationWithPubDirTest.java
```diff
@@ -41,8 +44,16 @@ public void testCreateWithPubDirSegment() throws NodeException {
...
-			NodeLoadResponse response = resource.add(request);
-			ContentNodeRESTUtils.assertResponseOK(response);
+			return resource.add(request);
+		});
+
+		ContentNodeRESTUtils.assertResponseOK(response);
+
+		Trx.operate(t -> {
+			com.gentics.contentnode.object.Node node = t.getObject(com.gentics.contentnode.object.Node.class, response.getNode().getId());
+			assertEquals(node.getName(), "Node name");
```
npomaroli, line 54: Those assertions are not descriptive. Better use assertThat(node).hasName(...).hasHostname(...).isNotHttps() and implement the missing methods in class NodeAssert.
The comment also applies to other tests (below).

npomaroli (changes requested):

npomaroli (approved):

PR #526, opened 2025-03-21
F gpu 1861

plyhun:


cms-core/src/main/resources/packages/DefaultElements/constructs/gtxalohapagelink/part.handlebars.hbs
```diff
@@ -1,3 +1,3 @@
 {{#gtx_alohalink cms.tag cms.rendermode ~}}
-<a{{{title}}}{{{target}}}{{{class}}}{{{data}}} href="{{{href}}}">{{cms.tag.parts.text.text}}</a>
+<a{{{title}}}{{{target}}}{{{class}}}{{{language}}}data}}} href="{{{href}}}">{{cms.tag.parts.text.text}}</a>
```
deckdom, line 2: The `data` is now not properly rendered.
Should be `{{{data}}}` and not `data}}}`

cms-core/src/main/resources/packages/DefaultElements/constructs/gtxalohapagelink/gentics_structure.json
```diff
@@ -152,6 +152,23 @@
...
     "externalEditorUrl" : "",
     "regexId" : 0
+  }, {
+    "globalId" : "A547.85233",
+    "name" : {
+      "de" : "Sprache",
+      "en" : "Language"
+    },
+    "keyword" : "language",
+    "editable" : false,
```
deckdom, line 162: Has to be `true` so we can edit it

cms-core/src/main/resources/packages/DefaultElements/handlebars/helpers/gtx_alohalink.js
```diff
@@ -29,6 +30,9 @@ function gtx_alohalink(tag, renderMode, options) {
     let href = '';
 
     if (url.target && url.target.id) {
+        if (url.language && url.language.code) {
+            ctx['language'] = ' hreflang="' + url.language.code + '"';
+        }
```
deckdom, line 35: here it has to load the language from the target, and not from the URL.
Additionally checks if language has already been set (If the user wants to override the language, then they should be able to).
```js
// Set the language only then, if no language has been saved yet.
if (!parts.language.text && url.target.language && url.target.language.code) {
    ctx['language'] = ' hreflang="' + url.target.language.code + '"';
}
```

deckdom (changes requested):
Added comments where things still have to be changed

deckdom (approved):

PR #533, opened 2025-04-02
F gpu 1908

plyhun:


plyhun:
Reopened from https://github.com/gentics/cms-oss/pull/525 .

PR #535, opened 2025-04-03
GPU-1877 Support user login

yrucrem:


cms-core/src/main/java/com/gentics/contentnode/rest/resource/impl/UserResourceImpl.java
```diff
@@ -558,6 +559,14 @@ public UserLoadResponse get(
 	public UserLoadResponse update(@PathParam("id") String id, User item) throws NodeException {
 		try (Trx trx = ContentNodeHelper.trx()) {
 			SystemUser user = MiscUtils.load(SystemUser.class, id, ObjectPermission.edit);
+
+			if (user.isSupportUser() && StringUtils.isNotEmpty(item.getPassword())) {
+				return new UserLoadResponse(
+					new Message(Type.WARNING, "password", new CNI18nString("rest.user.support_user_password").toString()),
```
npomaroli, line 565: Use I18nHelper.get() instead of new CNI18nString().toString()

PR #539, opened 2025-04-08
SUP-18372

deckdom:


cms-ui/libs/image-editor/src/lib/components/focal-point-selector/focal-point-selector.component.scss
deckdom, line 1: Correct styling of selected crosshair

PR #541, opened 2025-04-08
SUP-18376: Zero based option values for the select control

plyhun:


cms-ui/apps/editor-ui/src/app/tag-editor/components/tag-property-editors/select-tag-property-editor/select-tag-property-editor.component.ts
```diff
@@ -59,18 +59,7 @@ export class SelectTagPropertyEditor implements TagPropertyEditor {
     }
 
     onSingleSelectChange(newValue: number): void {
-        this.tagProperty.selectedOptions = [];
```
plyhun, line 62: Not directly related, but follows DRY.

deckdom (changes requested):

deckdom (approved):

PR #555, opened 2025-04-29
SUP-18442: Better Mesh Administration UX

plyhun:


cms-ui/libs/ui-core/src/lib/components/base-table/base-table.component.scss
```diff
@@ -136,6 +136,8 @@
 
     .cell-content-wrapper {
         margin: auto 0;
```
deckdom, line 138: Text has now a space in the beginning - Probably `pre` formatting

PR #557, opened 2025-05-07
SUP-18495: Added i18n-duration pipe; Fixed duration formatting; Fixed… 

deckdom:
… table styling


cms-ui/apps/admin-ui/src/app/shared/components/base-entity-table/base-entity-table.component.scss
```diff
@@ -13,18 +13,11 @@
     display: flex;
     flex-direction: row;
     flex-wrap: nowrap;
+    padding: 1rem 0;
+    gap: 0.5rem;
```
deckdom, line 17: Add flex-wrap on mobile

PR #570, opened 2025-06-03
SUP-18595: Fix the diff maker on HTML tags

plyhun:


cms-core/src/main/java/com/gentics/contentnode/rest/resource/impl/DiffResourceImpl.java
```diff
@@ -204,6 +229,51 @@ protected String renderDiff(DiffRequest request, boolean diffSource) throws Reso
...
+		int lastClosingTagPos = change.lastIndexOf("</");
+
+		if (firstOpeningTagPos > -1 && firstClosingTagPos > -1 && (firstOpeningTagPos+2) > firstClosingTagPos) {
+			int endPos = change.indexOf(" ", firstClosingTagPos+2);
+			if (endPos < 0) {
+				endPos = change.indexOf(">", firstClosingTagPos+2);
+			}
+			if (endPos > firstClosingTagPos+2) {
+				String tag = change.substring(firstClosingTagPos+2, endPos);
+				if (!"br".equals(tag) && !"img".equals(tag)) {
```
plyhun, line 257: Any other tags that allow no closing?

PR #640, opened 2025-09-22
Hotfix 6.3.x sup 18935

deckdom:


cms-ui/apps/editor-ui/src/app/core/components/actions-selector/actions-selector.component.html
```diff
@@ -51,14 +52,15 @@
                         </ng-template>
 
                         <ng-template #regularButton>
-                            <div
+                            <button
```
deckdom, line 55: Check if styling is still ok

PR #746, opened 2026-02-11
GPU-2196: OpenAPI generation tests

plyhun:
Technically there is nothing changed to the features or stability of the CMS, so no changelog.


cms-core/src/main/java/com/gentics/contentnode/rest/resource/impl/GroupResourceImpl.java
```diff
@@ -736,10 +736,10 @@ public TypePermissionList getPerms(@PathParam("id") String id, @QueryParam("pare
 
 	@Override
 	@GET
-	@Path("/{id}/perms/{type}")
+	@Path("/{id}/perms/{permType}")
```
npomaroli: Why is it necessary to rename this?

npomaroli (changes requested):

PR #792, opened 2026-05-05
Device-Preview

deckdom:


cms-ui/apps/editor-ui/src/app/content-frame/providers/device-preview/device-preview.service.ts
```diff
@@ -0,0 +1,166 @@
...
+        // browser back/forward, links).
+        this.routerSubscription = this.router.events.pipe(
+            filter((e): e is NavigationEnd => e instanceof NavigationEnd),
+        ).subscribe((event) => {
+            this.syncStateFromUrl(event.urlAfterRedirects);
+        });
+    }
+
+    public ngOnDestroy(): void {
+        this.routerSubscription?.unsubscribe();
```
deckdom: This is an invalid conditional chain.
`this.routerSubscription?.unsubscribe?.();` would be correct.
Use a simple `if` wrapper with a null check instead however.

PR #793, opened 2026-05-05
Content copilot UI

deckdom:


cms-ui/apps/editor-ui/src/app/copilot/providers/copilot-config/copilot-yaml.parser.ts
deckdom, line 1: We don't need a YAML Parser.
The Configuration itself should either be a JSON file, or in the future will be done in the CMS, and will be made available via a REST Endpoint in JSON.

cms-ui/apps/editor-ui/src/app/copilot/copilot.types.ts
```diff
@@ -0,0 +1,67 @@
...
+export interface CopilotAction {
+    /** Stable identifier, used as the Angular `trackBy` key. */
+    id: string;
+
+    /**
+     * Translation key OR literal label. Convention: if it contains a
+     * dot it's treated as an i18n key, otherwise rendered as-is. Keeps
+     * configuration files concise without precluding localisation.
+     */
+    label: string;
```
deckdom: Should be `labelI18n` with a `I18nString` type, as translation strings only work if they are bundled into the editor-ui already, which limits the use-case for this.

cms-ui/apps/editor-ui/src/app/copilot/copilot.types.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    /** Optional Material-icon name shown next to the label. */
+    icon?: string;
+
+    /**
+     * Short description shown under the label. Same dot-vs-literal
+     * heuristic as `label`.
+     */
+    description?: string;
```
deckdom: Same as `label` above, with `descriptionI18n`

cms-ui/apps/editor-ui/src/app/copilot/providers/copilot-config/copilot-yaml.parser.spec.ts
deckdom, line 1: Remove the tests for the parser, as there shouldn't be a parser to begin with

cms-ui/apps/editor-ui/src/app/copilot/providers/copilot-state/copilot-state.service.ts
deckdom, line 1: The state should be done in the already used `NgXS` store, instead of a custom solution.
Since it's only a single boolean flag, it should be integrated into the existing `ui` state.

cms-ui/apps/editor-ui/src/app/copilot/copilot.example.yml
deckdom, line 1: Needs to be a JSON for the UI to load. We won't include a YAML parser just for this.
Also move this to the root of the `cms-ui` or similar - This is the location for source files and not for example configuration files.

cms-ui/apps/editor-ui/src/app/copilot/copilot.types.ts
deckdom, line 1: Type definitions which are for configurations or for the REST API are to be put into the `cms-models` library.

cms-ui/apps/editor-ui/src/styles/_variables.scss
```diff
@@ -56,6 +56,12 @@ $gcms-form-editor-transition-duration: 0.15s;
 @mixin gcmsEditorBoxShadowFormElement {
     box-shadow: rgba( 0, 0, 0, 0.3 ) 0 0 14px -2px;
 }
+// Soft drop shadow for slide-in panels (Content Copilot drawer etc.).
+// Lighter than the form-element shadow because the host element is
+// flush against the edge of the workspace rather than floating.
+@mixin gcmsDrawerBoxShadow {
```
deckdom, line 62: This mixin is used once for the new sidebar.
Remove this mixin and use the style directly.

cms-ui/apps/editor-ui/src/app/copilot/providers/copilot-config/copilot-config.service.ts
```diff
@@ -0,0 +1,74 @@
...
+@Injectable({ providedIn: 'root' })
+export class CopilotConfigService {
+
+    private readonly configSubject = new BehaviorSubject<CopilotConfig>(DEFAULT_COPILOT_CONFIG);
+
+    /** The current configuration. Emits at least once with the default. */
+    public readonly config$: Observable<CopilotConfig> = this.configSubject.asObservable();
+
+    /** Convenience selector for the master enable flag. */
+    public readonly enabled$: Observable<boolean> = this.config$.pipe(
```
deckdom, line 35: If a feature is enabled is determined by the CMS Features/Copilot Endpoint.
The check for that should be done like any other Feature check instead of this individual solution.

cms-ui/apps/editor-ui/src/app/copilot/providers/copilot-config/copilot-config.service.ts
```diff
@@ -0,0 +1,74 @@
...
+        map((cfg) => !!cfg.enabled),
+    );
+
+    /** Convenience selector for the configured action list. */
+    public readonly actions$ = this.config$.pipe(map((cfg) => cfg.actions));
```
deckdom, line 40: The actions should be stored in the regular state of the editor-ui, using `NgXS`, to reuse existing patterns.
Especially since we don't reload configurations, using Observables is not useful.

cms-ui/apps/editor-ui/e2e/copilot.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+            await expect(copilotEmptyState(page)).toHaveCount(0);
+
+            const items = copilotSidebar(page).locator('.copilot-action-item');
+            await expect(items).toHaveCount(2);
+
+            const summarize = copilotActionItem(page, 'summarize');
+            await expect(summarize).toBeVisible();
+            await expect(summarize.locator('.copilot-action-label')).toHaveText('Zusammenfassen');
+            await expect(summarize.locator('.copilot-action-description'))
+                .toHaveText('Eine kurze Zusammenfassung der Seite erstellen');
```
deckdom: Checking for translatable text values is not a good way to test.
While the settings are stubed and therefore the values are known, these texts can't be directly used  due to changes for `label` -> `labelI18n` and `description` -> `descriptionI18n`.

We'd first need to ensure a specific UI language (as it's initially determined by the browser/system language), and therefore translation texts may differ.

cms-ui/apps/editor-ui/src/app/content-frame/components/editor-toolbar/editor-toolbar.component.html
```diff
@@ -140,6 +140,23 @@
...
+                        size="small"
+                        flat
+                        type="secondary"
+                        data-action="copilot"
+                        [class.is-active]="copilotOpen"
+                        [attr.data-active]="copilotOpen"
+                        [title]="'copilot.button_tooltip' | gtxI18n"
+                        (click)="toggleCopilot()"
+                    >
+                        <icon left>auto_awesome</icon>
```
deckdom: There's no icon named `auto_awesome` in the material symbols font.

cms-ui/apps/editor-ui/src/app/copilot/providers/copilot-config/copilot-config.service.ts
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Fetch and parse `{ui-conf}/config/copilot.yml`.
+     *
+     * Triggered once during application bootstrap. The cache-buster
+     * (`?t=`) follows the convention from `UIOverridesService` so a
+     * customer dropping a new file in does not have to wait for cache
+     * invalidation.
+     */
+    public load(): void {
+        const url = `${CUSTOMER_CONFIG_PATH}config/copilot.yml?t=${Date.now()}`;
```
deckdom: Why is the path `ui-conf/config`? Should be simply `ui-conf`

cms-ui/apps/editor-ui/src/app/copilot/components/copilot-sidebar/copilot-sidebar.component.html
```diff
@@ -0,0 +1,80 @@
...
+                    @for (action of actions; track trackById($index, action)) {
+                        <li
+                            class="copilot-action-item"
+                            [attr.data-id]="action.id"
+                        >
+                            @if (action.icon) {
+                                <icon class="copilot-action-icon">{{ action.icon }}</icon>
+                            }
+                            <div class="copilot-action-text">
+                                <span class="copilot-action-label">{{ action.label }}</span>
```
deckdom: The label and description need to be translated using the `gtxI18n` pipes, or with the changes for `labelI18n` and `descriptionI18n` using the `gtxI18nObject` pipe.

cms-ui/apps/editor-ui/src/app/copilot/components/copilot-sidebar/copilot-sidebar.component.html
```diff
@@ -0,0 +1,80 @@
...
+                        </li>
+                    }
+                </ul>
+            }
+        </div>
+    </section>
+
+    <footer class="copilot-sidebar-footer">
+        <div class="copilot-chat-input">
+            <textarea
```
deckdom: Use the existing `gtx-textarea` component instead and remove special styling for this textarea.

PR #808, opened 2026-05-26
GPU-2441

deckdom:


cms-ui/apps/form-translations/src/app/components/toasts/toasts.component.ts
deckdom, line 1: This should simply use the existing notification system from the `ui-core` instead of creating it's own again.

cms-ui/apps/form-translations/src/app/components/shell/shell.component.ts
deckdom, line 1: Why does this component exist?
Initialization should be done in the angular `provideAppInitializer` if possible, or in the `app.component.ts`, as that one is currently pretty much empty.

cms-ui/apps/form-translations/src/app/core/services/notification.service.ts
deckdom, line 1: Use the `I18nNotificationService` from the `cms-components` package instead of making your own.

cms-ui/apps/form-translations/src/app/core/services/tool-api.service.ts
```diff
@@ -0,0 +1,118 @@
+import { Injectable, NgZone } from '@angular/core';
+import { BehaviorSubject, Observable, Subject } from 'rxjs';
+
+import { environment } from '../../../environments/environment';
+
+/**
+ * Nachrichten Tool → CMS-Shell.
+ */
+type OutgoingMessage =
```
deckdom: Typings are wrong - Use the types actually defined from `cms-integration-api-models`.
In general the `ct-link-checker` also has a tool api -> This should be consolidated to the `cms-components` package instead.

cms-ui/apps/form-translations/src/app/core/services/authentication.service.ts
```diff
@@ -0,0 +1,79 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject, Observable } from 'rxjs';
+
+import { environment } from '../../../environments/environment';
+
+const STORAGE_KEY = 'gtx-form-translations.sid';
```
deckdom: It should try to load the SID from `GCMSUI_sid` instead - This is what all apps use.

cms-ui/apps/ct-form-translations/src/app/core/services/authentication.service.ts
deckdom, line 1: Also implemented in the `ct-link-checker`.
Should be consolidated into `cms-components` as well.

cms-ui/apps/form-translations/src/app/core/services/app.service.ts
deckdom, line 1: This whole service is only used for initialization, which could just as well be done in the component as well and doesn't need this extra state handling.

cms-ui/apps/form-translations/src/app/pipes/translate.pipe.ts
deckdom, line 1: Dont create a custom translation pipe. Use the existing `gtxI18n` pipe.

cms-ui/apps/form-translations/src/app/services/form-translations.service.ts
deckdom, line 1: This is the state management for translations - Why is it not called anything that indicates that this is handling the state?
Why is there a custom state solution instead of using the already used `ngxs` store?

Is there a reason why we need this state in the first place?
This seems like it can easily be implemented using regular component states instead of creating a own global state.

cms-ui/apps/form-translations/src/app/shared/translation.service.ts
deckdom, line 1: Don't use a custom translation service/handling. Use the existing translation mechanics from `cms-components`/`ngx-translate`

cms-ui/apps/form-translations/src/app/services/form-translations.service.ts
```diff
@@ -0,0 +1,431 @@
...
+    await this.loadScope(activeScopeId);
+  }
+
+  /* ---------- Laden ---------- */
+
+  private async loadLanguages(): Promise<void> {
+    this.patch({ languagesLoading: 'loading' });
+    try {
+      const dto = await firstValueFrom(
+        this.http.get<LanguagesResponseDto>(`${environment.apiBaseUrl}/form/translations/languages`)
```
deckdom: Don't call APIs like this directly.
Add the correct types into `cms-models`, and add the API to the `cms-rest-client` instead.

cms-ui/apps/ct-form-translations/src/styles/styles.scss
deckdom, line 1: Why is the entire styling so completely different from the rest of the Editor and Admin UI?
This doesn't look well integrated if it looks and works completely different from the existing solutions.

cms-ui/apps/form-translations/cms-registration-config.json
deckdom, line 1: 1. This is an example for the CMS configuration, so it should be in `yaml` instead, as the CMS config is not in json.
2. The definition is wrong, as the `id` is missing and the properties `group`, and `description` don't exist. https://www.gentics.com/Content.Node/cmp8/guides/admin_custom_tools.html#gsc.tab=0. It'd also be good to have the name be a translatable object instead, to support multiple languages.

cms-ui/apps/form-translations/src/app/models/dto.model.ts
deckdom, line 1: Models from the API (Without any business logic) are to be added to the `cms-models` package instead.
Use existing models instead of re-defining everything manually.

cms-ui/apps/form-translations/src/app/components/translations-table/translations-table.component.html
deckdom, line 1: A certain reason why the `gtx-table` hasn't been used? Would help a lot with styling consistency

cms-ui/apps/ct-form-translations/src/app/components/save-bar/save-bar.component.ts
deckdom, line 1: Why is the save bar at the bottom? We have the save button in the top right in both the editor- and admin-ui.

cms-ui/apps/form-translations/src/app/components/save-bar/save-bar.component.html
```diff
@@ -0,0 +1,38 @@
...
+    {{ 'COMMON.DISCARD' | gtxTranslate }}
+  </button>
+  <button
+    type="button"
+    class="pm-btn pm-btn-primary"
+    data-action="save"
+    [disabled]="dirtyCount === 0 || isSaving"
+    (click)="save.emit()">
+    <span *ngIf="isSaving" class="pm-spinner pm-spinner--sm"></span>
+    <svg *ngIf="!isSaving" width="16" height="16" viewBox="0 0 24 24" fill="none"
```
deckdom: Use material symbols instead of hardcoding some SVGs out of nowhere

cms-ui/apps/ct-form-translations/e2e/scope-switch.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+        });
+
+        await test.step('Common Test Setup', async () => {
+            await IMPORTER.cleanupTest();
+            await IMPORTER.syncPackages(TestSize.MINIMAL);
+            await IMPORTER.setupTest(TestSize.MINIMAL);
+        });
+
+        await test.step('Open tool', async () => {
+            const sid = String(IMPORTER.client?.sid);
```
deckdom, line 46: This login doesn't work properly, as it's trying to use the SID from the importer without having access to the cookies.
Either login with a separate client in the page context to properly set the cookie, or add a auth flow liek the editor/admin-ui with a login field.

cms-ui/apps/ct-form-translations/e2e/common.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+/**
+ * Default credentials for tests that don't need a custom user setup.
+ *
+ * Prefer creating a dedicated test user via the EntityImporter where possible
+ * (see the integration-test guide). Use this admin only for smoke tests where
+ * permissions aren't the thing under test.
+ */
+export const AUTH = {
```
deckdom, line 15: Shouldn't use the admin user for tests - Create a dedicated user with adequate permissions for each test instead.

PR #812, opened 2026-06-09
SUP-19869: Rearrange language status icons, remove optional display

plyhun:


cms-ui/apps/editor-ui/src/app/list-view/components/item-list/item-list.component.ts
```diff
@@ -103,7 +103,7 @@ export class ItemListComponent implements OnInit, OnChanges, OnDestroy {
     icon = '';
     showGrid = true;
 
-    public displayStatusInfo: boolean;
+    public displayStatusInfo = true;
```
plyhun, line 106: I did not dare removing these too, as they look like being connected to the other properties of the components around.

PR #814, opened 2026-06-16
SUP-19848: Copypaste fix

plyhun:


cms-core/src/main/java/com/gentics/contentnode/publish/protocol/PublishProtocolUtil.java
```diff
@@ -61,7 +61,7 @@ private static void updatePublishStateOfNodeObject(
 		var tableName = getType(object.getTType()).toLowerCase();
 
 		if (PublishState.ONLINE.getValue() == logEntry.getState()) {
-			var publisherId = object.getFutureUnpublisher() != null ? object.getFutureUnpublisher().getId() : userId;
+			var publisherId = object.getFuturePublisher() != null ? object.getFuturePublisher().getId() : userId;
```
plyhun, line 64: Been a bit confused of how this has to behave actually. Like, if the object is to be (un)published right away, its scheduled state (future(Un)publisher) has to be cleared away in advance.

PR #817, opened 2026-06-18
SUP-19921: Fix the scheduled wastebin purge of a referenced page

plyhun:


cms-core/src/main/java/com/gentics/contentnode/etc/LangTrx.java
```diff
@@ -135,11 +135,7 @@ protected void setCode() throws NodeException {
 		DBUtils.executeStatement("SELECT short FROM language WHERE id = ?", new SQLExecutor() {
 			@Override
 			public void prepareStatement(PreparedStatement stmt) throws SQLException {
```
plyhun, line 137: Note from `TransactionManager::getLanguageId()`
```
				// TODO: actually, we should always and only use the
				// session's languageId. Calling
				// ContentNodeHelper.getLanguageId() is done for backwards
				// compatibility, so that we don't change any current
				// behaviour. This will actually throw expected exceptions because
				// ContentNodeHelper.getLanguageId() isn't used by
				// ContentNodeResource. To fix this, one should remove the
				// get/setLanguageId() from ContentNodeHelper at some point.
				// Another point: this transaction may actually be for user=0
				// and sessionId=null (that happens when viewing a page
				// it seems). In that case, ContentNodeHelper.getLanguageId()
				// may still have a valid languageId - it should be checked whether
				// that is so, because then, one _has_ to use
				// ContentNodeHelper.getLanguageId() since the languageId can't
				// be retrieved if the sessionId is null.
```

Some of the current usages of `ContentNodeHelper.getLanguageId()` belongs to the wraps of UserLanguageFactory.getById()`, which looks like working the no-session case correclty. Other usages are somewhat unclear, whether expecting an exception on this method, so I hesitate so far removing it completely, as per comment above.

PR #822, opened 2026-06-25
SUP-19876: Fix the regression of missing content staging status icons

plyhun:


cms-ui/apps/editor-ui/e2e/list-loading.spec.ts
```diff
@@ -297,4 +298,48 @@ test.describe('List Loading', () => {
             }
         });
     });
+
+
+    test.describe('Content Staging', () => {
```
plyhun, line 304: As usual, the UT won't work until the fix is released and packed into the container.

deckdom: You can run the tests locally, with full UI to test with `npm run e2e editor-ui -- --ui`

cms-ui/apps/editor-ui/e2e/list-loading.spec.ts
```diff
@@ -297,4 +298,48 @@ test.describe('List Loading', () => {
...
+
+            await page.locator('gtx-content-staging-modal gtx-content-package-list gtx-contents-list-item gtx-checkbox').click();
+            await page.locator('gtx-content-staging-modal button.primary[data-action="primary"]').click();
+
+            await page.locator('gtx-button.staging-mode-leaver').isVisible();
+            const folderList = findList(page, 'folder');
+            const folderItems = await folderList.locator('item-list-row').all();
+            for (const item of folderItems) {
+                await expect(item.locator('.status-label icon')).toBeVisible();
+                await expect(item.locator('.status-label')).toHaveText('Nicht beinhaltet');
```
deckdom: Never test against text values like this, as this is a translation string, and would break if the language is different.
Already the case, since on default the tests will be run in English as UI language.

Use classes or other identifiers to determine the state.
If you need some/they aren't present, simply add them.

PR #827, opened 2026-07-02
Hotfix 6.3.x sup 19628

plyhun:


cms-core/src/test/java/com/gentics/contentnode/tests/rest/group/GroupEditTest.java
```diff
@@ -28,17 +37,27 @@
...
 	 * Test creating a new group (as subgroup of node group)
 	 * @throws NodeException
 	 */
+	@Parameter(0)
+	public int userId;
+
+	@Parameters(name = "{index}: user ID {0}")
+	public static Collection<Object[]> data() {
+		// 1 for system, 3 for super admin
+		return List.of(new Object[] { 1 }, new Object[] { 3 });
```
plyhun, line 54: I am a bit worried that most of our data manipulation tests use `system` group, and not the `Super Admin` one, whose possibilities are limited.

cms-ui/apps/admin-ui/e2e/group.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+                expect(actionButtons.locator('.table-action-button button.btn').filter({ hasText: 'Neuen Benutzer in Gruppe erstellen' })).not.toBeVisible();
+                expect(actionButtons.locator('.table-action-button button.btn').filter({ hasText: 'Benutzer zuweisen' })).not.toBeVisible();
+            });
+
+            await test.step('Node Super Admin group should allow creating a subgroup', async () => {
+                const tab = page.locator(SELECTORS.TAB_GROUP_SUBS);
+                await tab.click();
+                const actionButtons = page.locator('gtx-group-detail gtx-tab[data-id="subgroups"] .entity-table-actions-bar');
+                await actionButtons.waitFor();
+                expect(actionButtons.locator('.table-action-button button.btn').filter({ hasText: 'Neue Untergruppe erstellen' })).toBeVisible();
```
plyhun, line 90: Yes, I do remember using data-* instead of text content. Somehow they are not working for my setup, as expected :(

cms-ui/apps/admin-ui/src/app/core/providers/operations/group/group.operations.ts
```diff
@@ -93,7 +96,8 @@ export class GroupOperations extends ExtendedEntityOperationsBase<'group'> {
      * Get a single group and add it to the AppState.
      */
     get(userId: number): Observable<Group<Raw>> {
-        return this.api.group.getGroup(userId).pipe(
+        return this.api.group.getGroup(userId, { perms: true }).pipe(
+            tap((res: GroupResponse) => this.permissions.storePermissions(AccessControlledType.GROUP_ADMIN, res.group.id, res.perms || GcmsPermission.VIEW)),
```
plyhun, line 100: This obviously won't work without backend changes.

cms-restapi/src/main/java/com/gentics/contentnode/rest/resource/parameter/FilterParameterBean.java
```diff
@@ -12,8 +16,19 @@ public class FilterParameterBean {
 	@QueryParam("q")
 	public String query;
 
+	/**
+	 * Query string for permissions filter
+	 */
+	@QueryParam("permitted")
+	public List<Permission> permitted;
```
npomaroli: Do not add this here. The FilterParameterBean is used in lots of list endpoints, which do not support filtering for permissions. By adding this here, this parameter would be documented for all list endpoints, not only the one (list groups) which actually supports this.

cms-core/src/test/java/com/gentics/contentnode/tests/rest/group/GroupEditTest.java
npomaroli: This is because data manipulation tests and permission check tests are separated.

cms-core/src/test/java/com/gentics/contentnode/tests/rest/group/GroupEditTest.java
```diff
@@ -28,17 +37,27 @@
...
+	}
+
 	@Test
 	public void testCreate() throws NodeException {
 		String name = "New Group";
 		String description = "Group description";
 		GroupResourceImpl resource = new GroupResourceImpl();
-		Group created = Trx.supply(() -> {
+		Group created = Trx.supply(Trx.supply(trx -> trx.getObject(SystemUser.class, userId)), () -> {
```
npomaroli: Load the system user only once (e.g. in setup of the test), not every time a new transaction is created,

npomaroli (changes requested):

cms-ui/apps/admin-ui/src/app/features/group/components/group-detail/group-detail.component.ts
```diff
@@ -168,4 +168,14 @@ export class GroupDetailComponent extends BaseDetailComponent<'group', GroupOper
         this.fgProperties.markAsPristine();
     }
 
+    protected isGroupAssignable(): Observable<boolean> {
```
deckdom: This check doesn't seem to be correct/complete.
With the node user I can still see the buttons - Not sure if it's because it requires the backend changes, but from what I saw in the changes this doesn't seem to be the case.

Don't use functions for bindings if possible.
Create a simple observable for this like this instead:
```ts
public isGroupAssignable$: Observable<boolean>;

ngOnInit(): void {
    // ...
    this.isGroupAssignable$ = this.currentEntity$.pipe(
        switchMap((entity) => {
            if (!entity) {
                return of(false);
            }

            return this.permissionsService.getPermissions(AccessControlledType.GROUP_ADMIN, entity.id).pipe(
                map(typePermissions => {
                    return typePermissions.hasPermission(GcmsPermission.USER_ASSIGNMENT);
                }),
            );
        }),
        publishReplay(1),
        refCount(),
    );
}
```

npomaroli (changes requested):
Please redirect to hotfix-6.4.x and add a changelog entry

PR #830, opened 2026-07-08
GPU-2599 update node to v26.4.0

dominikschatz:
GPU-2599 update node to v26.4.0


deckdom:
Will be resolved with #859 instead.

PR #843, opened 2026-07-23
SUP-20110: Mesh dep update

plyhun:
No changelog because of no actual change aside of CMP component update, which is already documented.


npomaroli:
This MR is not necessary as we will raise the Mesh Version when releasing the CMS anyways.
Setting the version to a mesh version, which does not exist right now will cause the build to fail.

PR #852, opened 2026-07-29
SUP-19852: Added e2e test and fixed visual inconsistencies

deckdom:


cms-ui/apps/editor-ui/src/app/content-frame/components/aloha-color-picker-renderer/aloha-color-picker-renderer.component.ts
deckdom, line 1: When text is colored, doesn't detect the initial color / doesn't show it in the color picker (library, not in the palette)

PR #855, opened 2026-08-05
GPU-2631 UI: Add UI for Api Token Management

dominikschatz:
GPU-2631 UI: Add UI for Api Token Management


cms-ui/apps/admin-ui/e2e/api-tokens.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+    test('can save when form is valid', async ({ page }) => {
+        await openApiTokenModal(page);
+        
+        await openApiTokenCreateModal(page);
+
+        const form = page.locator(API_CREATE_MODAL).locator("form");
+
+        await expect(form).toBeVisible();
+
+        const submitBtn = page.locator(API_CREATE_MODAL).locator(".modal-footer").locator("button.primary");
```
deckdom: Please add action attributes (like `data-action="confirm"`) to the buttons like other modals.
For confirming the modal, please use the `clickModalAction` helper then.

cms-ui/apps/admin-ui/e2e/api-tokens.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+        const tomorrow = new Date();
+        tomorrow.setDate(tomorrow.getDate() + 1);
+        
+        await dateInput.fill(tomorrow.toISOString().split('T')[0]);
+
+        await expect(submitBtn).not.toBeDisabled();
+
+        await submitBtn.click();
+
+        await expect(page.locator(".success-message")).toBeVisible();
```
deckdom: Please add an `id` to the success-notification (`NotificationService/I18nNotificationService.show`), and then use the `findNotification` helper to get the notification by the ID.

cms-ui/apps/admin-ui/e2e/api-tokens.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+
+        const row = table.locator(".data-row").first();
+
+        const deleteButton = row.locator('[title="editor.tagtype_delete_label"]').first();
+
+        await deleteButton.click();
+
+        await expect(page.locator(API_DELETE_MODAL)).toBeVisible();
+
+        await page.locator(API_DELETE_MODAL).locator("button.alert").click();
```
deckdom: Use actions here as well

cms-ui/apps/admin-ui/e2e/api-tokens.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+    test('can delete Api Tokens', async ({ page }) => {
+        await openApiTokenModal(page);
+        
+        const table = page.locator(API_MODAL).locator("gtx-manage-api-tokens-table");
+
+        await expect(table).toBeVisible();
+
+        const row = table.locator(".data-row").first();
+
+        const deleteButton = row.locator('[title="editor.tagtype_delete_label"]').first();
```
deckdom: Use the `findTableAction` helper instead, and provide the delete action-id instead.

cms-ui/apps/admin-ui/e2e/api-tokens.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+        
+        const table = page.locator(API_MODAL).locator("gtx-manage-api-tokens-table");
+
+        await expect(table).toBeVisible();
+
+        const row = table.locator(".header-row");
+
+        await row.locator('gtx-checkbox').click();
+
+        await row.locator('button.alert').click();
```
deckdom: Use the `findTableAction` helper

cms-ui/apps/admin-ui/e2e/api-tokens.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+
+        await expect(page.locator(API_DELETE_MODAL)).toBeVisible();
+
+        await page.locator(API_DELETE_MODAL).locator("button.alert").click();
```
deckdom: action-attribute

cms-ui/apps/admin-ui/e2e/api-tokens.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+
+        await expect(table.locator(".data-row")).toHaveCount(0);
+    });
+    
+});
+
+async function openApiTokenModal(page: Page) {
+    await page.locator(".ng-trigger-toggleState").click();
```
deckdom: Don't use angular internal `ng` prefixed classes for selectors.
Especially not `trigger` classes, as these are for animations.

Use the `openContext` function instead (See changes for the user-menu` first):
```ts
const menu = page.locator('gtx-user-menu');
const dropdown = await openContext(menu.locator('.user-name gtx-dropdown-list'));
await dropdown.locator('[data-action="manage-api-tokens"]').click();
```

cms-ui/libs/cms-components/src/lib/components/user-menu/user-menu.tpl.html
```diff
@@ -33,9 +33,10 @@
     <div class="user-details">
       <div class="user-name">
         <div>
-          <span [innerHTML]="'user.logged_in_as' | gtxI18n:{ name: getUserName() }"></span>
+          <span [innerHTML]="'user.logged_in_as' | gtxI18n"></span>
           <gtx-dropdown-list>
```
deckdom: Add `data-context-id` attributes to the `gtx-dropdown-list`. and `gtx-dropdown-content` elements (same ID), and the `data-context-trigger` .
This allows for the `openContext` helper to properly detect and relate the elements together.

cms-ui/libs/cms-components/src/lib/components/user-menu/user-menu.tpl.html
```diff
@@ -46,6 +47,9 @@
               <gtx-dropdown-item overrideSlot="changePasswordOption" (click)="showPasswordModalClicked()">
                 {{ 'user.change_password_label' | gtxI18n }}
               </gtx-dropdown-item>
+              <gtx-dropdown-item overrideSlot="manageApiTokensOption" (click)="showApiTokensModalClicked()">
```
deckdom: Do not add a overrideSlot here.
Add a `data-action` attribute for later use instead.

cms-ui/apps/admin-ui/e2e/api-tokens.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+    await page.locator("gtx-dropdown-list gtx-button").last().click();
+    await page.locator('[overrideslot="manageApiTokensOption"]').click();
+}
+
+async function openApiTokenCreateModal(page: Page) {
+    const modalTable = page.locator(API_MODAL).locator("gtx-manage-api-tokens-table");
+    const createNewBtn = modalTable.locator(".entity-table-actions-bar button");
```
deckdom: Add `data-action` attribute

cms-ui/apps/admin-ui/e2e/api-tokens.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+    await createNewBtn.click();
+}
+
+async function clearEntries(IMPORTER: EntityImporter) {
```
deckdom: Create a dedicated test-user in the setup instead - See most of the editor-ui tests.
This removes the need to cleanup these tokens and we can do permission checks.

cms-ui/apps/admin-ui/src/app/app.component.ts
```diff
@@ -222,6 +223,13 @@ export class AppComponent implements OnDestroy, OnInit {
             .catch(this.errorHandler.catch);
     }
 
+    onShowApiTokenModal(): void {
```
deckdom: Why do we loop around into the app?
Since everything is already contained in the `cms-components` library, the modal can be simply opened from the `user-menu` component directly.

cms-ui/libs/cms-components/src/lib/components/api-tokens-create-form/api-tokens-create-form.component.html
deckdom, line 1: Don't nest the components, we use a flat hierarchy.
i.E. move all `components/api-token/*` to `components/*`.

Also, for this component, call it `api-token-properties` instead, so it has the same naming as all other components that are forms/properties.

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-create-form/api-tokens-create-form.component.ts
```diff
@@ -0,0 +1,105 @@
...
+        generateValidatorProvider(ApiTokensCreateFormComponent),
+    ],
+    standalone: false,
+})
+export class ApiTokensCreateFormComponent extends BaseFormPropertiesComponent<EditableApiTokenPackage> {
+    @Input()
+    private readonly apiTokenNames: Array<string>;
+
+    @Output()
+    onValidChange = new EventEmitter<boolean>();
```
deckdom: Don't use the `on` prefix for outputs.
Also, we don't need this output, as the parent should have a `FormControl` connected to this component anyways, which handles all of this already.

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-create-form/api-tokens-create-form.component.ts
```diff
@@ -0,0 +1,105 @@
...
+    
+    @Output()
+    onCreated = new EventEmitter<string | null>();
```
deckdom: This is also redundant, as this component doesn't create the token, it's only for managing the form/property data.

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-create-form/api-tokens-create-form.component.ts
```diff
@@ -0,0 +1,105 @@
...
+        });
+    }
+
+    protected override configureForm(value: EditableApiTokenPackage, loud?: boolean): void {}
+
+    protected override assembleValue(value: EditableApiTokenPackage): EditableApiTokenPackage {
+        return value;
+    }
+
+    public addApiToken(): void {
```
deckdom: This component should not create the token.
That's the job of the modal, this component is only here to manage the form/property data of a token.

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-create-form/api-tokens-create-form.component.ts
```diff
@@ -0,0 +1,105 @@
...
+        this.handler.create(submitData)
+            .subscribe({
+                next: (newToken) => {
+                    this.onCreated.emit(newToken.token);
+                },
+                complete: () => this.onCreated.emit(null)
+            });
+    }
+
+    private readonly futureDateValidator: ValidatorFn = (control: AbstractControl): ValidationErrors | null => {
```
deckdom: Since this function doesn't refer to the component directly, move it out of the class, maybe even into the `ui-core/utils/forms`, as this seems like a rather useful basic validator to use.

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-create-modal/api-tokens-create-modal.component.html
```diff
@@ -0,0 +1,44 @@
+<div class="modal-content">
+  @if(createdToken() !== null) {
+    <h2>{{ 'api_token.token_created' | gtxI18n }}</h2>
+    <div class="success-message">
+      <p>{{ 'api_token.token_store' | gtxI18n }}</p>
+      <p>
```
deckdom: Why the random `<p>`?

cms-ui/libs/cms-components/src/lib/components/api-tokens-create-modal/api-tokens-create-modal.component.scss
```diff
@@ -0,0 +1,16 @@
...
+        padding-top: 10px;
+        text-align: center;
+
+        .copy-icon {
+            vertical-align: bottom;
+            cursor: pointer;
+        }
+    }
+
+    gtx-button.loading {
```
deckdom, line 12: On load simply disable the button instead, as this is what all the modals do.

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-create-modal/api-tokens-create-modal.component.ts
```diff
@@ -0,0 +1,57 @@
...
+@Component({
+    selector: 'gtx-api-tokens-create-modal',
+    templateUrl: './api-tokens-create-modal.component.html',
+    styleUrls: ['./api-tokens-create-modal.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    standalone: false,
+})
+export class ApiTokensCreateModalComponent implements IModalDialog {
+    @ViewChild(ApiTokensCreateFormComponent)
+    readonly createForm!: ApiTokensCreateFormComponent;
```
deckdom: Don't access the child component, especially not to call functions within it like this.
If you need to do something like this, then the architecture of the components is plain wrong.

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-create-modal/api-tokens-create-modal.component.ts
```diff
@@ -0,0 +1,57 @@
...
+
+
+@Component({
+    selector: 'gtx-api-tokens-create-modal',
+    templateUrl: './api-tokens-create-modal.component.html',
+    styleUrls: ['./api-tokens-create-modal.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    standalone: false,
+})
+export class ApiTokensCreateModalComponent implements IModalDialog {
```
deckdom: Extend the `BaseModal` class instead, which handles the registration already

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-create-modal/api-tokens-create-modal.component.ts
```diff
@@ -0,0 +1,57 @@
...
+    registerCloseFn(close: (val?: any) => void): void {
+        this.closeFn = close;
+    }
+    
+    registerCancelFn(cancel: (val?: any) => void): void {
+        this.cancelFn = cancel;
+    }
+
+    public closeModal() {
+        this.createdToken.set(null);
```
deckdom: We don't need to reset this - The modal is getting closed anyways

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-delete-modal/api-tokens-delete-modal.component.html
deckdom, line 1: This entire modal is redundant - Use the `ModalService.dialog` function instead to open such a basic modal.
The deletion of the token then should be moved to the table component.

cms-ui/libs/cms-components/src/lib/components/api-tokens-modal/api-tokens-modal.component.html
deckdom, line 1: Use `api-token-modal.component.html` as name - `tpl` is only from old legacy components

cms-ui/libs/cms-components/src/lib/components/api-tokens-modal/api-tokens-modal.component.html
```diff
@@ -0,0 +1,11 @@
+<div class="modal-content">
+  <gtx-manage-api-tokens-table></gtx-manage-api-tokens-table>
+</div>
+<div class="modal-footer">
+  <gtx-button
+    type="secondary" flat
```
deckdom, line 6: Add `data-action` attribute

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-table/api-tokens-table.component.ts
```diff
@@ -0,0 +1,166 @@
...
+    selector: 'gtx-manage-api-tokens-table',
+    templateUrl: './api-tokens-table.tpl.html',
+    styleUrls: ['./api-tokens-table.component.scss'],
+    standalone: false
+})
+export class ApiTokensTableComponent implements OnInit {
+    public readonly TableSelectAllType = TableSelectAllType;
+
+    @Input()
+    isDisabled: boolean;
```
deckdom: Don't use `is` as prefix - simply call it `disabled`

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-table/api-tokens-table.component.ts
```diff
@@ -0,0 +1,166 @@
...
+                id: 'valid',
+                label: this.i18n.instant('license.status_VALID'),
+                fieldPath: 'valid',
+            },
+        ];
+
+        this.refreshData();
+    }
+
+    private refreshData(): void {
```
deckdom: Where's the pagination?

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-table/api-tokens-table.component.ts
```diff
@@ -0,0 +1,166 @@
...
+})
+export class ApiTokensTableComponent implements OnInit {
+    public readonly TableSelectAllType = TableSelectAllType;
+
+    @Input()
+    isDisabled: boolean;
+
+    public readonly apiTokenRows = signal<TableRow<ApiTokenBo>[]>([]);
+    
+    protected apiTokenColumns: TableColumn<ApiTokenBo>[] = [];
```
deckdom: Why are the rows the only ones which are signals?
Also please update the naming of the variables - It's quite obvious we're talking about `api-tokens` in the `api-tokens-table`, i.E. call them `rows`, `columns`,  ...

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-create-form/api-tokens-create-form.component.ts
```diff
@@ -0,0 +1,105 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    providers: [
+        generateFormProvider(ApiTokensCreateFormComponent),
+        generateValidatorProvider(ApiTokensCreateFormComponent),
+    ],
+    standalone: false,
+})
+export class ApiTokensCreateFormComponent extends BaseFormPropertiesComponent<EditableApiTokenPackage> {
+    @Input()
+    private readonly apiTokenNames: Array<string>;
```
deckdom: Inputs can't be private

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-table/api-tokens-table.component.ts
```diff
@@ -0,0 +1,166 @@
...
+        })
+        .then(modal => modal.open())
+        .catch((error) => console.log(error));
+    }
+
+    public updateSelection(selections: Array<string>) {
+        this.apiTokenSelection = selections;
+    }
+
+    private toDateString(dateNumber: number): string {
```
deckdom: Use the `gtxI18nDate` pipe for date formatting instead.
Especially since this is hardcoded to use `de` as language.
Must be done with a custom renderer in the template, just like booleans.

cms-ui/libs/cms-components/src/lib/components/api-tokens/api-tokens-table/api-tokens-table.component.ts
```diff
@@ -0,0 +1,166 @@
...
+        if(dateNumber === 0) {
+            return "-";
+        }
+
+        return new Intl.DateTimeFormat('de-DE').format(
+            new Date(dateNumber * 1000)
+        );
+    }
+
+    public getApiTokenNames() : Array<String> {
```
deckdom: Unused, remove

cms-ui/libs/cms-components/src/lib/components/copy-value/copy-value.component.ts
deckdom, line 1: Rename the component selector to `gtx-copy-value`, as `gtx-mesh-copy-value` doesn't make sense.

cms-ui/libs/cms-components/src/lib/components/user-menu/user-menu.component.ts
```diff
@@ -89,4 +91,8 @@ export class UserMenuComponent {
     showPasswordModalClicked(): void {
         this.showPasswordModal.emit();
     }
+
+    showApiTokensModalClicked(): void {
+        this.showApiTokensModal.emit();
```
deckdom: Open the modal directly here, no point to loop out to the apps

cms-ui/libs/cms-components/src/lib/components/user-menu/user-menu.component.ts
```diff
@@ -73,6 +73,8 @@ export class UserMenuComponent {
     @Output() toggle = new EventEmitter<boolean>();
     /** On event user clicks showPasswordModal button. */
     @Output() showPasswordModal = new EventEmitter<void>();
+    /** On event user clicks showPasswordModal button. */
+    @Output() showApiTokensModal = new EventEmitter<void>();
```
deckdom: To be removed

cms-ui/libs/cms-components/src/lib/components/user-menu/user-menu.tpl.html
```diff
@@ -33,9 +33,10 @@
     <div class="user-details">
       <div class="user-name">
         <div>
-          <span [innerHTML]="'user.logged_in_as' | gtxI18n:{ name: getUserName() }"></span>
+          <span [innerHTML]="'user.logged_in_as' | gtxI18n"></span>
```
deckdom: Revert and move the `span` inside the `gtx-dropdown-trigger`

cms-ui/libs/cms-components/src/lib/components/user-menu/user-menu.tpl.html
```diff
@@ -33,9 +33,10 @@
...
           <gtx-dropdown-list>
             <gtx-dropdown-trigger>
+              <strong>{{getUserName() + "'"}}</strong>
               <a>
```
deckdom: remove the `a` element - we don't need that

cms-ui/libs/cms-models/src/lib/api-token.ts
```diff
@@ -0,0 +1,30 @@
+
+/**
+ * An Api Token object as returned from the /admin/token endpoint:
+ */
+
+export interface ApiTokenBo {
```
deckdom: Do not add BO (Business-Object) to the `cms-models`.
This package is only supposed to contain the actual Rest-API models.
Business logic has to be defined in the library or application layer.

cms-ui/libs/cms-models/src/lib/api-token.ts
```diff
@@ -0,0 +1,30 @@
...
+    id: number;
+    userId: number;
+    name: string;
+    cdate: string;
+    expires: string;
+    lastUsed: string;
+    valid: boolean;
+}
+
+export type EditableApiTokenPackage = Pick<TokenData, 'name' | 'expires'>;
```
deckdom: What is a `ApiTokenPackage`??
Call it `EditableApiTokenData`

cms-ui/libs/cms-models/src/lib/request.ts
```diff
@@ -2503,3 +2504,16 @@ export interface PushLicenseRequest {
...
+
+// API TOKENS /////////////////////////////////////////////////////////////////////////////////////////////////////
+
+export interface ApiTokenListOptions extends BaseListOptionsWithPaging<ApiToken> {}
+
+/**
+ * Request object used to add an api token for the
+ * `/adnim/token` endpoint.
+ */
+export interface ApiTokenCreateRequest {
```
deckdom: Use the `EditableApiTokenData` which we defined earlier, i.E. 

`export interface ApiTokenCreateRequest extends EditableApiTokenData {}`

cms-ui/libs/cms-models/src/lib/response.ts
```diff
@@ -1533,3 +1533,28 @@ export interface KeycloakConfiguration {
     'resource': string;
     'showSSOButton': boolean;
 }
+
+// API TOKENS /////////////////////////////////////////////////////////////////////////////////////////////////////
+
+export interface ApiTokenListResponse extends ListResponse<ApiTokenResponse> { }
+
+export interface ApiTokenResponse extends Response {
```
deckdom: Why are all fields re-defined??
We have the model already defined, reuse them.

cms-ui/libs/cms-models/src/lib/api-token.ts
```diff
@@ -0,0 +1,30 @@
...
+    lastUsed: string;
+    valid: boolean;
+}
+
+export interface ApiToken {
+    token: string;
+    data: TokenData
+}
+
+interface TokenData {
```
deckdom: Call it `ApiTokenData`

cms-ui/libs/cms-models/src/lib/response.ts
```diff
@@ -1533,3 +1533,28 @@ export interface KeycloakConfiguration {
...
+    id: number;
+    userId: number;
+    name:string;
+    cdate: number;
+    expires: number;
+    lastUsed: number;
+    valid: boolean;
+}
+
+export interface ApiTokenData {
```
deckdom: Delete this - This is duplicated *again* and is the same as the `TokenData` from the other file.

cms-ui/libs/cms-components/src/lib/providers/api-token-handler/api-token-handler.service.ts
deckdom, line 1: Remove this service - The cms-components doesn't provide yet another abstraction for the rest-client.
Simply use the rest-client in the modal/table directly, as it's only 2 endpoints anyways that we use.

cms-ui/libs/cms-rest-clients-angular/src/lib/api-token/api-token-api.ts
deckdom, line 1: Delete this - The `cms-rest-clients-angular` library is deprecated, and we'll not add new functionality to it, as we want to get rid of it.

deckdom (changes requested):
Generally please check all the formatting of the `ts` filles.
VSCode lights up like a christmas tree on every file.
ESLint should be able to fix most if not all formatting issues anyways.

cms-ui/libs/cms-components/src/lib/components/api-tokens-create-form/api-tokens-create-form.component.ts
deckdom, line 1: Rename to `api-tokens-properties.component.ts` (and html/scss file)

cms-ui/libs/cms-components/src/lib/components/api-tokens-create-form/api-tokens-create-form.component.ts
```diff
@@ -0,0 +1,73 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    providers: [
+        generateFormProvider(ApiTokensCreateFormComponent),
+        generateValidatorProvider(ApiTokensCreateFormComponent),
+    ],
+    standalone: false,
+})
+export class ApiTokensCreateFormComponent extends BaseFormPropertiesComponent<EditableApiToken> {
+    @Input()
+    control: FormControl<EditableApiToken>;
```
deckdom, line 19: Remove this - This component doesn't need the control, as this control is the `form`.

The structure should look like this:
- Modal Component
-- Has a `control` property (i.E. this property here)
-- Template then contains `<gtx-api-tokens-properties [formControl]="control" />`

The Modal component can now check the validity of the entire form via `control.valid` and get the value via `control.value`.

cms-ui/libs/cms-components/src/lib/components/api-tokens-create-form/api-tokens-create-form.component.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    @Input()
+    readonly apiTokenNames: Array<string>;
+
+    @Output()
+    validChange = new EventEmitter<boolean>();
```
deckdom, line 25: Remove this as well

cms-ui/libs/cms-components/src/lib/components/api-tokens-create-form/api-tokens-create-form.component.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    constructor(
+        changeDetector: ChangeDetectorRef,
+    ) {
+        super(changeDetector);
+    }
+
+    override ngOnInit(): void {
```
deckdom, line 33: Entire override can be caned as well, as the two extra subscriptions aren't needed

cms-ui/libs/cms-components/src/lib/components/api-tokens-create-modal/api-tokens-create-modal.component.html
```diff
@@ -1,32 +1,31 @@
...
     <div class="success-message">
       <p>{{ 'api_token.token_store' | gtxI18n }}</p>
-      <p>
-        <gtx-mesh-copy-value
-            class="copy-api-token"
-            [value]="createdToken()"
-            [animate]="true"
-        />
-      </p>
+      <gtx-copy-value
```
deckdom, line 7: The modal itself should only do one thing - House the properties component and create the new token.
This keeps the logic all grouped together and we can reuse more parts.
When the token was created, simply close the modal and open a dedicated `copy-value-modal`  instead.
Note: The component which opened the `api-tokens-create-modal` should handle the opening of the `copy-value-modal`.

cms-ui/libs/cms-components/src/lib/components/api-tokens-create-modal/api-tokens-create-modal.component.html
```diff
@@ -1,32 +1,31 @@
...
+          class="copy-api-token"
+          [value]="createdToken()"
+          [animate]="true"
+      />
     </div>
   } @else {
     <h2>{{ 'api_token.new' | gtxI18n }}</h2>
     <gtx-api-tokens-create-form
```
deckdom, line 15: Why do we have the `gtx-token-properties` *and* this form in here?
Does this component even exist?
Don't double bind a form-control, as this just causes issues.
And the `validChange` isn't needed as the state is already handled via the `control`.

cms-ui/libs/cms-components/src/lib/components/api-tokens-create-modal/api-tokens-create-modal.component.ts
```diff
@@ -0,0 +1,87 @@
...
+import { I18nNotificationService } from '../../providers';
+
+@Component({
+    selector: 'gtx-api-tokens-create-modal',
+    templateUrl: './api-tokens-create-modal.component.html',
+    styleUrls: ['./api-tokens-create-modal.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    standalone: false,
+})
+export class ApiTokensCreateModalComponent implements IModalDialog {
```
deckdom, line 15: Extend the `BaseModal` please.
Generic should be the response from the token create (or just the token string).

cms-ui/libs/cms-components/src/lib/components/api-tokens-create-modal/api-tokens-create-modal.component.ts
```diff
@@ -0,0 +1,87 @@
...
+        };
+
+        this.loading.set(true);
+
+        this.api.admin.addApiTokens(submitData)
+            .subscribe({
+                next: (newToken) => {
+                    const name = newToken.data.name;
+                    this.createdToken.set(newToken.token);
+                    console.log(newToken);
```
deckdom, line 69: Please remove logging of the token - Probably just left over from debuging.

cms-ui/libs/cms-components/src/lib/components/api-tokens-modal/api-tokens-modal.component.html
deckdom: Use the generic `close` action like all other modals please.

cms-ui/libs/cms-components/src/lib/components/api-tokens-modal/api-tokens-modal.component.ts
```diff
@@ -1,15 +1,14 @@
...
+import { IModalDialog } from '@gentics/ui-core';
 
 @Component({
     selector: 'gtx-api-tokens-modal',
-    templateUrl: './api-tokens-modal.tpl.html',
-    standalone: false
+    templateUrl: './api-tokens-modal.component.html',
+    standalone: false,
 })
 export class ApiTokensModalComponent implements IModalDialog {
```
deckdom, line 9: Extend the `BaseModal<void>` please

cms-ui/libs/cms-components/src/lib/components/api-tokens-table/api-tokens-table.component.html
```diff
@@ -33,4 +45,9 @@ <h2 class="item-tag-list-heading">{{ 'api_token.tokens' | gtxI18n }}</h2>
     class="gtx-table-cell-boolean"
     [ngClass]="'gtx-table-cell-boolean-' + value"
   >{{ value == null ? '' : (value ? 'check' : 'close') }}</icon>
+</ng-template>
+<ng-template #dateRenderer let-value="value">
+  <span class="gtx-table-cell-text">
+    {{ value | gtxI18nDate:'date' }}
```
deckdom, line 51: Should be date-time, so we can see the time info as well (As usually tokens run out/are created to a specfic time, and not just "00:00").
```suggestion
    {{ value | gtxI18nDate:'dateTime' }}
```

cms-ui/libs/cms-components/src/lib/components/api-tokens-table/api-tokens-table.component.ts
```diff
@@ -0,0 +1,216 @@
+import { Component, OnInit, signal } from '@angular/core';
+import { ModalService, TableAction, TableActionClickEvent, TableColumn, TableRow, TableSelectAllType } from '@gentics/ui-core';
+import { forkJoin, map } from 'rxjs';
+import { GCMSRestClientService } from '@gentics/cms-rest-client-angular';
+import { I18nNotificationService, I18nService } from '../../providers';
+import { ApiTokensCreateModalComponent } from '../api-tokens-create-modal/api-tokens-create-modal.component';
+
+interface ApiTokenBo {
```
deckdom, line 8: Why do we need a BO for the tokens in this Table?
The regular Rest-Model should just work fine

cms-ui/libs/cms-components/src/lib/components/api-tokens-table/api-tokens-table.component.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    private refreshData(): void {
+        const sort = this.sortOrder() === 'asc' ? `+${this.sortBy()}` : `-${this.sortBy()}`;
+        this.client.admin.getApiTokens(this.perPage, this.page(), sort)
+            .pipe(
+                map((res) => {
+                    this.totalCount.set(res.numItems);
+                    return res.items.map((token) => ({
+                        id: `${token.id}`,
+                        item: {
```
deckdom, line 105: Please avoid item mapping if we don't need it (i.E. BO mapping).
For simple value mappings/changes, you can use the `mapper` property in the `column` definition if needed (For the dates for example).

As far as I know, the `gtxI18nDate` pipe/formatter should be able to detect the timestamps correctly (i.E. the `* 1000` might not be needed).

cms-ui/libs/cms-models/src/lib/api-token.ts
```diff
@@ -1,30 +1,24 @@
...
-    lastUsed: string;
+    cdate: number;
+    expires: number;
+    lastUsed: number;
     valid: boolean;
 }
 
-export type EditableApiTokenPackage = Pick<TokenData, 'name' | 'expires'>;
\ No newline at end of file
+export type EditableApiToken = {
```
deckdom, line 21: Please use type inheritance for this:

```ts
export interface EditableApiToken {
  name: string;
  expires?: string;
}

export interface ApiTokenData extends EditableApiToken {
  readonly cdate: number;
  // etc.
}
```

cms-ui/libs/cms-models/src/lib/api-token.ts
```diff
@@ -1,30 +1,24 @@
...
-
 export interface ApiToken {
     token: string;
-    data: TokenData
+    data: ApiTokenData;
 }
 
-interface TokenData {
+export interface ApiTokenData {
+    token: string;
```
deckdom, line 11: The token value is only visible in the `ApiTokenCreateResponse`, therefore please remove it from here and add it to the Response-Model instead.

cms-ui/libs/cms-rest-client/src/lib/abstracts.ts
```diff
@@ -417,8 +418,8 @@ export interface AbstractAdminAPI extends BasicAPI {
     getPublishQueue: () => PublishQueue;
     modifyPublishQueue: (body: ContentMaintenanceActionRequest) => Response;
 
-    getApiTokens: () => ListResponse<ApiTokenResponse>;
-    addApiTokens: (body: ApiTokenCreateRequest) => ApiTokenResponse;
+    getApiTokens: (pageSize?: number, page?: number, sort?: string) => ListResponse<ApiTokenResponse>;
```
deckdom, line 421: Rename it to `listApiTokens` instead, to make the name consistent with the naming of the other functions (at least as well as we can).
Create a `ApiTokenListOptions` Model in the `request.ts` with these properties instead and provide it as object instead, like all other "list" functions do.

cms-ui/libs/cms-rest-client/src/lib/client.ts
```diff
@@ -223,7 +223,7 @@ export class GCMSRestClient implements GCMSRootAPI {
         getPublishQueue: () => this.executeMappedJsonRequest(GET, '/admin/content/publishqueue'),
         modifyPublishQueue: (body) => this.executeMappedJsonRequest(POST, '/admin/content/publishqueue', body),
 
-        getApiTokens: () => this.executeMappedJsonRequest(GET, '/admin/token'),
+        getApiTokens: (pageSize: number = 10, page: number = 1, sort: string = '-cdate') => this.executeMappedJsonRequest(GET, `/admin/token?pageSize=${pageSize}&page=${page}&sort=${sort}`),
```
deckdom, line 226: Don't simply append the options yourself. There's an extra argument you may pass, i.E. 
`this.executeMappedJsonRequest(GET, '/admin/token', null, params)`

cms-ui/libs/ui-core/src/lib/utils/forms.ts
```diff
@@ -206,3 +206,17 @@ function asRegExp(pattern: string | RegExp): null | { str: string; regex: RegExp
 
     return { str, regex };
 }
+
+export function futureDateValidator(control: AbstractControl): ValidationErrors | null {
```
deckdom, line 210: This should probably have an extra option if the time should be cleared.
On default we shouldn't clear the time (i.E. line 219), as otherwise we would allow someone to create a token which is already expired (or whereever this validator is also going to be used in the future)

deckdom (changes requested):

PR #858, opened 2026-08-11
Hotfix 6.4.x sup 19935

plyhun:


deckdom:
Changes are good - But the tests are missing. Please add them, then we can merge it.

